Issue #39, opened 2025-08-28
handle breaking changes


lhaibach:
We want to explore extending the page classification with new classes, e.g geo_profile, table, diagram - and assess how easily each classifier (baseline, pixtral, layoutlmv3, treebased) can be extended. While this work is in progress, prediction outputs must remain compatible with the application.

PR #40, opened 2025-08-28
Feat/issue 39/breaking changes handling

lhaibach:
This PR introduces configuration for multiple output profiles following the extension of ground truth + classes in #39. It adds **prediction profiles** that control which labels are emitted and how any extended labels are mapped, so the output remains compatible with the app (see: apps/server-asset-sg/src/features/assets/files/file-processors/file-extraction.service.ts in the swissgeol-assets-suite repository)

@daniel-va , @Tugark 
We’d like to standardize class names to singular, lower_case for consistency (e.g., text, map, title_page, boreprofile, unknown, geo_profile, table, diagram).
If it’s easy for you to switch on your side, we’ll keep this convention. If not I will map them to the exact naming and spelling as in the app.

- added a prediction profile mechanism (`PREDICTION_PROFILE`) to `.env.template`, allowing choice between 'stable' and 'dev' output modes for classification results.
- `stable` : allows for text, map, title_page, boreprofile and unknown and `dev`: allows additionally geo_profile, table and diagrams.


tests/test_api_mapping.py
```diff
@@ -0,0 +1,26 @@
...
+            {
+                "page": 1,
+                "classification": cls or {"text": 1, "boreprofile": 0, "map": 0, "title_page": 0, "unknown": 0},
+                "metadata": {"language": "de", "is_frontpage": False},
+            }
+        ],
+    }
+
+
+def test_mapping_renames_v1_keys_only():
```
stijnvermeeren-swisstopo, line 18: "v1" is not used anymore elsewhere, so best to rename this test method as well.

stijnvermeeren-swisstopo (approved):
I added comments for a few small things I noticed, but it's nothing critical, so I'm already "pre-approving" 👍

PR #44, opened 2025-09-16
Feat/issue 28/baseline geo profiles

### LINKED ISSUE #28: Extend class detection to geo_profiles
We want to extend the baseline classifier with logic to detect **Geological Profiles**. These are geological cross-sections or longitudinal profiles that differ from typical boreprofiles.

Characteristic features of such pages:
- **Geometric lines**: long connected lines are common (often horizontal/vertical, but also many non-vertical lines).
- **Connected boreprofiles**: multiple boreprofiles may be shown, with layers linked via geometric lines.
- **Keywords:** terms like längsprofil, geologisches profil.
- **Legends / scale bars**: presence of a Maßstab or explanatory legend.
- **Orientation markers**: compass directions (e.g., NW, SE, SW).
- **Layout:** often span large parts of the page.
- **Text**: text density of pages is low and high sparsity
- **Colour**: often the areas are coloured , several colours per page
Features we calculate can then be introduced into treebased classifier as well in a follow up.

TicaGit:
Resolves #28 

This pull request introduces new detection functionality for geological cross-sections and profiles.

The algorithm leverages the presence of specific keywords and reuses the line score method from map detection, but applies a lower threshold.

A new color analysis module is added to extract, discretize, and filter colors found in documents. While this feature was not useful for baseline geoprofile detection, it could be valuable for the Treebase classifier to improve classification accuracy.

Sidebar-based heuristics (e.g., "more than 2 sidebars = geo profile") cannot be used reliably because sidebars are not consistently detected (small numbers are not picked up by OCR). Filtering is also insufficient (random numeric columns are incorrectly identified as borehole sidebars without correction, validation, or association with material descriptions).

As the number of new classes increases, fine-tuning thresholds and selecting the right features by hand for classification become more difficult. This explains the low baseline accuracy achieved.

| Metric       | text  | boreprofile | map   | **geo_profile** | title_page | diagram | table | unknown | Macro Avg |
|--------------|-------|-------------|-------|-------------|------------|---------|-------|---------|-----------|
| F1 Score (%) | 65.00 | 75.47       | 74.51 | **36.36**       | 50.00      | 0.00    | 0.00  | 21.95   | 40.41     |

Suggestions to Improve the Tree-Based Classifier
- Color features: Use the number of colors or their proportions as input features to help distinguish document types.
- Sidebar features: Replace the boolean has_sidebar with a numeric num_sidebar, so the tree can decide its own threshold instead of being limited to true/false.

Note:
- the flag PREDICTION_PROFILE must be set to "dev" in .env to also classify pages into the new classes (like geo profiles).


lhaibach:
Thanks a lot for investigating the different approaches and for the detailed description. I agree that manually fine-tuning thresholds and features doesn’t make sense. But still we can use the newly added features in the tree-based models, and assess whether the color module contributes meaningfully to the model’s decisions f.e :)

PR #46, opened 2025-09-18
Detect diagrams

lhaibach:
This pull request introduces logic for detecting diagrams. The diagram detection is based on keyword matching and identifying x and y axis scales. 

### Diagram detection

* Added a new `identify_diagram` to detect diagram pages based on keywords, units, and the presence of numeric axes. Numeric axis: identifed via clustering numeric entries on x and y position, values have to be mostly in or decreasing to be considered a scale. (via `is_mostly_increasing` function)
* Updated `config/matching_params.yml` to add `diagram` keywords in multiple languages and a list of units.
### sidebar detection
* Replaced the strict (`is_strictly_increasing`) with a more tolerant `is_mostly_increasing` function in `src/identifiers/boreprofile.py`, which allows up to 20% violations when determining if a sequence is mostly increasing. This makes boreprofile detection more robust to noise.

|Metric | text | boreprofile | map | geo_profile | title_page | diagram | table | unknown | Macro Avg
|-- | -- | -- | -- | -- | -- | -- | -- | -- | --
|F1 Score (%) | 65.00 | 77.77 | 73.07 | 0.00 | 48.78 | 0.55 | 0.00 | 24.66 | 43.04

Currently detection of diagrams does not leverage geometric lines, which might improve the scores. This could further be investigated, also the detection of table like structures (as in the boreholes repository).

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -322,3 +322,66 @@ title_page:
   it:
     - rapporto
     - progetto
+
+diagram:
+  en:
+    - diagram
+    - curve
+    - distribution
+    - histogram
+    - summation curve
+    - cumulative curve
+    - grading curve
+  de:
+    - kurve
+    - diagramm
+    - kornverteilung
+    - histogramm
+    - summationskurve
+    - körnungslinie
+    - verteilungskurve
+  fr:
+    - diagramme
+    - courbe
+    - distribution
+    - histogramme
+    - courbe de sommation
+    - courbe granulométrique
+  it:
+    - diagramma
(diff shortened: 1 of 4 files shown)


lhaibach:
changing the detection logic from 
```
if has_keywords and has_unit:
     return True
...
return y_ok and x_ok
```

to 
```
votes = sum([has_keyword, has_unit, y_ok, x_ok])
return votes >=2
```

improves metrics to diagram: F1=62.22%, Precision=60.87%, Recall=63.64%, but might be too loose?

Issue #47, opened 2025-09-18
Detect Tables for Baseline Classifier


lhaibach:
Add logic to detect tables for baseline classifier.

Tables usually contain:
- text that is structured in y aligned columns
- often geometric grid lines / table like structures
-> if table like structure is present then their is structured text inside (compared to diagrams where we also have table like structures, but usually not structured text)

Next to an initial logic for the baseline classifier, this issue aims to calculate new features that can be used to extend TreeBased Model. Cumbersome finetuning and feature selection is out of the scope of this ticket.

PR #54, opened 2025-09-24
table detection via word positioning

lhaibach:
Roseolves #47 This pull request adds table detection to the baseline classifier using logic solely based on text alignment and structure. 
**Table detection and classification:**

* Added `src/identifiers/table.py` that implements functions to detect tables on a page based on confidence about a detected table (grouped TextColumns).

**extended text object model:**

* Added new dataclasses `TextColumn` and `TextTable` to `src/text_objects.py`, representing vertical columns and tables of text, respectively. These classes have properties for bounding box merging, text coverage, and table confidence (based on row alignment)
* Refactored `cluster_text_elements` to accept a float instead of integers
* Created generic BFS function cluster_connected_components in `src/text_objects.py` which is used both in creation of clusters in title page detection and table detection
**Metrics:**

|Metric | text | boreprofile | map | geo_profile | title_page | diagram | table | unknown | Macro Avg
|-- | -- | -- | -- | -- | -- | -- | -- | -- | --
|F1 Score (%) | 65.00 |75.47 | 74.51 |36.36 |50.00| 0.00 |34.78 | 23.19 | 44.91

### DIFF
--- a/src/bounding_box.py
+++ b/src/bounding_box.py
@@ -16,6 +16,16 @@ def merge_bounding_boxes(rects):
     return pymupdf.Rect(x0, y0, x1, y1)
 
 
+def _x_center(rect: pymupdf.Rect) -> float:
+    """Finds middle x position of a rectangle."""
+    return 0.5 * (rect.x0 + rect.x1)
+
+
+def _y_center(rect: pymupdf.Rect) -> float:
+    """Finds middle y position of a rectangle."""
+    return 0.5 * (rect.y0 + rect.y1)
+
+
 def is_line_below_box(line_rect: pymupdf.Rect, image_rect: pymupdf.Rect) -> bool:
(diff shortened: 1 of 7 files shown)


lhaibach:
@TicaGit thanks for checking and pointing that out! I originally though building from the beginning of the page would be a good ideal, and just reused the function but you are right, that way for tables in text pages i will get "wrong" columns, missing the correct ones! I,ll change the function to `cluster_connected_components` and write a connecting function that checks x and y positioning of the words. So on y axis we have some min/ max distance that is allowed and make it a bit more forgiving (or adjust the way of checking x positioning), if i see im too strict there

TicaGit (approved):
I think its already quite nice, thanks for the changes. I saw that the table I mentioned in my previous comment is now separated from the little text above it, so it definitely improved. I think we can merge it like this, and refine it if we have other ideas.

PR #56, opened 2025-09-30
Feat/issue 48/extend tree features

TicaGit:


src/models/treebased/basetrainer.py
```diff
@@ -91,12 +91,15 @@ def evaluate(self, y_pred):
...
-        return {"precision_micro": precision, "recall_micro": recall, "f1_micro": f1}
+        _, _, f1_macro, _ = precision_recall_fscore_support(self.y_val, y_pred, average="macro", zero_division=0)
+        return {"precision_micro": precision, "recall_micro": recall, "f1_micro": f1, "f1_macro": f1_macro}
 
     def save_model(self, filename: str = "model.joblib"):
         """Saves the trained model to the specified file."""
         path = self.model_dir / filename
         joblib.dump(self.model, path)
+        signature = mlflow.models.infer_signature(self.X_train[:1], self.model.predict(self.X_train[:1]))
+        mlflow.sklearn.log_model(self.model, name=self.model_name, signature=signature)
```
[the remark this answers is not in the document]
TicaGit: this is wrong, it's deprecated

pyproject.toml
```diff
@@ -17,14 +17,15 @@ dependencies = [
...
-    "regex>=2024.11.6,<2025.0.0",
+    "regex==2025.7.34",
     "scipy==1.14.1",
     "tabulate>=0.9.0,<0.10.0",
     "tqdm>=4.67.1,<5.0.0",
     "pyyaml>=6.0.2,<7.0.0",
     "shapely>=2.1.1,<3.0.0",
     "scikit-image>=0.25.2,<0.26.0",
     "scikit-learn>=1.7.1,<2.0.0",
+    "scikit-optimize==0.10.2",
```
stijnvermeeren-swisstopo, line 28: Why is this required?

stijnvermeeren-swisstopo (changes requested):
I'm struggling with the training script.

If I run `python src.models.treebased.train.py` as mentioned in the README, I get "python: can't open file '/home/stijn3/swissgeol-assets-dataextraction/src.models.treebased.train.py': [Errno 2] No such file or directory"

If I run `python src/models/treebased/train.py` then instead I get     
```
Traceback (most recent call last):
  File "/home/stijn3/swissgeol-assets-dataextraction/src/models/treebased/train.py", line 14, in <module>
    from classifiers.pdf_dataset_builder import build_filename_to_label_map
  File "/home/stijn3/swissgeol-assets-dataextraction/venv/lib/python3.10/site-packages/classifiers/pdf_dataset_builder.py", line 10, in <module>
    from src.page_classes import label2id
ModuleNotFoundError: No module named 'src'
```

What am I doing wrong?

stijnvermeeren-swisstopo (changes requested):
Shouldn't we also update the "published" model that is contained in this repository under `models/stable/model.joblib` / `models/stable/xgboost/model.joblib` ? (I guess one of those must be there by mistake? In https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/57/, the second one is deleted, so I guess that answers that question.) 

I think it's also about time that we start looking into a better integration of the swissgeol-boreholes-dataextraction into this one. I guess the most logical path would be to have the swissgeol-boreholes-dataextraction as a dependency in this project? Then we could reuse some of the code from there without copy-pasting lots of code, as we currently do in e.g. `table_detection.py` or `geometric_objects.py`. Moreover, that would make it easier to benefit from ongoing improvements such as the extraction of borehole names (https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/300) that are also relevant to the data extraction from assets. Shall we create a new issue for this?

pyproject.toml
letao: I have improved the README.md file with new instructions and removed the unnecessary scikit-optimize and scikit-image packages

PR #57, opened 2025-10-02
refactor/issue-45/api-response-pydantic

TicaGit:
The `/collect` endpoint now returns a CollectResponse object instead of a JSONResponse. If processing fails, it raises a 500 error rather than returning a 200 status code.

CollectResponse follows essentially the same structure as the previous JSON dictionary:
```
{
    "has_finished": True,
    "data": [
        {
            "filename": "input.pdf",
            "metadata": {
                "page_count": 1,
                "languages": [
                    "fr"
                ]
            },
            "pages": [
                {
                    "predicted_class": "Map",
                    "page_number": 1,
                    "page_metadata": {
                        "language": "fr",
                        "is_frontpage": True
                    },
                }
            ],
        }
    ],
}
```
Compared to earlier responses, some fields have been renamed for clarity (e.g., page → page_number). Additionally, classification has been replaced by predicted_class, which now contains a simple string with the predicted value instead of a one-hot encoded dictionary of all classes. This change could be reverted if needed.

Previous schema for reference:
```
{
    "has_finished": False,
    "data": [
        {
            "filename": "sample.pdf",
            "metadata": {
                "page_count": 1,
                "languages": [
                    "de"
                ]
            },
            "pages": [
                {
                    "page": 1,
                    "classification": {
                        "text": 1,
                        "boreprofile": 0,
                        "map": 0,
                        "title_page": 0,
                        "unknown": 0
                    },
                    "metadata": {
                        "language": "de",
                        "is_frontpage": False
                    },
                }
            ],
        },
    ]
}
```
The class mapping has been updated to follow a consistent format.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -9,10 +9,11 @@ ENABLE_COLOR_PROPORTION=false
 AWS_MODEL_ID="your_model_id"
 AWS_MODEL_REGION="anthropic_region"
 
-FASTTEXT_MODEL_PATH = "path_to_FastText_model"
+FASTTEXT_MODEL_PATH="path_to_FastText_model"
 
 S3_BUCKET="asset-sg"
 S3_FOLDER="asset-sg/"
+TMP_PATH="/tmp"
 
 
 # set USE_LOCAL to True and the local configs to use a local S3 instance
--- a/api/api.py
+++ b/api/api.py
@@ -6,14 +6,17 @@ import uuid
 from pathlib import Path
 from typing import Annotated
 
-from aws import aws
 from fastapi import BackgroundTasks, Depends, FastAPI, HTTPException, Response, status
-from pydantic import BaseModel, Field
-from starlette.responses import JSONResponse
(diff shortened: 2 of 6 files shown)


stijnvermeeren-swisstopo (changes requested):
It looks to me that in the currently deployed version, there is not yet any `/v1/` prefix for the endpoints. As such, moving the current endpoints to a `/v1/` prefix is in itself a breaking change, that we should avoid. I would propose to keep the current endpoints (with the "old" schema) available under the current paths for now.
Apart from that, I'm quite happy with the new solutions with different API versions. I don't have feedback from EBP about this yet, but I think we can merge it after fixing the endpoint paths, and then create a follow-up ticket if there are any requirements from the web application side that we did not consider yet.

We should also briefly document the differences between the API versions in the README. (And how to run the API in the first place should be documented as well, that seems to be completely missing from the README.)

PR #61, opened 2025-10-17
Docker build improvements

letao:
The following improvements to the docker build setup:

- Created a proper `__init__.py` file structure to better handle package imports in the api. 
- Updated the docker build script in `dockerfile` and `pyproject.toml` to utilize only the packages that are needed for the runtime.
- Improved the copy command in the`dockerfile`  to only copy the necessary files for the runtime 
- Improved the `README.md` to better reflect the docker build commands

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ __pycache__/
 
 # Virtual environments
 env/
-.venv/
+venv/
 
 # IDE config
 .idea/


README.md
```diff
@@ -149,6 +149,7 @@ Requirements: Python 3.10(recommended), OCR'ed PDFs.
 ```bash
 python -m venv venv
 source venv/bin/activate
+pip install -e '.[all]'
```
TicaGit: if this is here, point 2. must be remove

christianabbet: I was wondering if .gitignore ahould be updated to exclude venv/ and not env/

README.md
```diff
@@ -151,20 +151,21 @@ python -m venv venv
 source venv/bin/activate
```
christianabbet: Adding venv in gitignore instead of env?

PR #65, opened 2025-10-28
implement model explainability

### LINKED ISSUE #62: Implementation of model interpretability features for XGBoost
Now that [swissgeol-assets-dataextraction](https://github.com/swisstopo/swissgeol-assets-dataextraction)  repository uses XGBoost as the main classification model, we should introduce explainability tools to better understand model predictions.

The goal is to identify which features contribute most to:

- Global importance: overall model behavior across all instances
- Local importance: specific predictions for individual instances

Two suitable, model-agnostic libraries are:

- [SHAP](https://shap.readthedocs.io/en/latest/) - widely used, strong community support
- [LIME](https://lime-ml.readthedocs.io/en/latest/) - flexible, interpretable outputs

Implement an ML explainability pipeline that generates and stores key interpretability metrics, enabling analysis of both global and local feature contributions for the XGBoost model.

TicaGit:
Resolves #62 

This PR implements model explainability capabilities for tree-based classifiers using SHAP (SHapley Additive exPlanations) values. The implementation allows users to generate visualizations that explain both individual predictions and overall model behavior.

Key changes:
- Added SHAP-based explanation functions for generating local waterfall and force plots and global beeswarm plot
- Added command-line flag (-x) to enable explanation plot generation during inference


src/classifiers/treebased_classifier.py
```diff
@@ -2,6 +2,7 @@
 
 import pymupdf
 
+from models.treebased.model_explanation import explain_prediction
```
stijnvermeeren-swisstopo: Should this be `from src.models...` ?

TicaGit:
I was able to reproduce the issue after reinstalling the Python virtual environment.

The problem comes from a breaking change introduced in XGBoost v3.1.x, which is incompatible with the latest SHAP release. SHAP expects the model intercept (base_score) to be a single float value, but newer versions of XGBoost store one value per class (eight in our case). Since our current model was trained with an older version, the single value 0.5 was duplicated across classes, leading to the reported error.

Reverting to XGBoost v3.0.5, which still uses a single float for the intercept, resolved the issue and restored compatibility with SHAP.

We should also make a habit of pinning dependency versions (avoiding loose constraints like > or <) to avoid similar compatibility issues in the future.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #70, opened 2025-11-03
hotfix/issue-69/mlflow-imports

### LINKED ISSUE #69: Dataextraction service fails due to missing dependencies
The current `edge` version fails on DEV due to missing dependencies:

```
Traceback (most recent call last):
  File "/app/api/utils/task.py", line 58, in run
    value = target()
  File "/app/api/app/shared/handlers.py", line 32, in <lambda>
    task.start(payload.file, background_tasks, lambda: process(payload, aws_client, settings))
                                                       ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/app/v1/app.py", line 109, in process
    result = script(
        input_path=tmp_dir,
    ...<2 lines>...
        write_result=False,
    )
  File "/app/main.py", line 100, in main
    classifier = create_classifier(classifier_type, model_path, matching_params, explain_model)
  File "/app/src/classifiers/classifier_factory.py", line 43, in create_classifier
    from src.classifiers.treebased_classifier import TreeBasedClassifier
  File "/app/src/classifiers/treebased_classifier.py", line 8, in <module>
    from src.models.treebased.model_explanation import explain_prediction
  File "/app/src/models/treebased/model_explanation.py", line 5, in <module>
    import mlflow
ModuleNotFoundError: No module named 'mlflow'
```

I believe this is because dependencies have been moved into subgroups (`mlflow` is now in `dev` dependencies) but our Docker image does not reflect these changes. Since I don't know if there are more runtime dependencies, I did not fix it myself, for example, also `torch` is only a `dev` dependency.

TicaGit:
Resolves #69 

This PR fixes MLflow import handling by making the import conditional, allowing optional MLflow tracking.

Currently, MLflow is a mandatory dependency for training tree-based models. To make training possible without MLflow, we would need to refactor the code further. While this may not be necessary right now, since model training is only performed by developers, automatic or scheduled training in the future could reintroduce dependency issues if MLflow isn’t available.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ dependencies = [
     "regex==2025.7.34",
     "scikit-learn>=1.7.1,<2.0.0",
     "shapely==2.1.2",
-    "shap==0.49.1",
     "scipy==1.14.1",
     "tabulate>=0.9.0,<0.10.0",
     "tqdm>=4.67.1,<5.0.0",
@@ -44,6 +43,7 @@ dev = [
     "matplotlib>=3.10.3,<4.0.0",
     "mlflow>=3.3.1,<4.0.0",
     "pytest-cov>=6.2.1,<7.0.0",
+    "shap==0.49.1",
     "torch>=2.7.1,<3.0.0",
     "transformers>=4.54.0,<5.0.0",
 ]
--- a/src/models/treebased/model_explanation.py
+++ b/src/models/treebased/model_explanation.py
@@ -1,12 +1,10 @@
 import io
+import logging
 import os
 import tempfile
 
 import numpy as np
-import shap
 from dotenv import load_dotenv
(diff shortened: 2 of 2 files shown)


letao (approved):
Looks good to me to address the underlying issue.

One small note, currently we are dependent on the MLFLOW_TRACKING=False variable for the API to function properly, in case we install runtime dependencies with 'pip install .' and set MLFLOW_TRACKING=True we will have errors with mlfow import in all the scripts where we have this line:
```Python
if mlflow_tracking:
    import mlflow
```

Ideally we would want package installs depend on an environment as opposed to an environment variable. @TicaGit, can we create a ticket that would aim to address together with creating a CI step that tests application behavior based on Production only dependencies in the pyproject.toml file.

PR #75, opened 2025-11-12
Language detection from boreholes package

### LINKED ISSUE #63: Update the language detection approach for assets  and boreholes
After completion of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/310 we should unify the language detection approach for both assets and boreholes repositories. Currently:

- **swissgeol-assets-dataextraction** uses [fasttext-predict · PyPI](https://pypi.org/project/fasttext-predict/)
A more lightweight fork of [facebookresearch/fastText: Library for fast text representation and classification.](https://github.com/facebookresearch/fastText), which itself is now archived. The fork itself has not been updated since November 2024.
- **swissgeol-boreholes-dataextraction** uses [langdetect · PyPI](https://pypi.org/project/langdetect/)
Based on an algorithm that dates back to 2024, and the Python library itself not been updated since 2021

Based on the teams discussion [here](https://teams.microsoft.com/l/message/19:39mTEQ7CM2bo-AQfrIpmWaILCIKUCirRbNRUby3bIEI1@thread.tacv2/1760354152828?tenantId=6ae27add-8276-4a38-88c1-3a9c1f973767&groupId=fa6d4b3b-be14-45d7-92d1-685c3ffee262&parentMessageId=1760354152828&teamName=GeologAI%20-%20Externe&channelName=General&createdTime=1760354152828)  we should look at an option to use substitute **fasttext** with **fast-langdetect** which is built on the same components but has regular updates and a lite version which might help reducing the memory needs for our APIs: [LlmKira/fast-langdetect: ⚡️ 80x faster Fasttext language detection out of the box | Split text by l…](https://github.com/LlmKira/fast-langdetect).

As par of this issue we should evaluate  **fast-langdetect**  to see if we can achieve the same performance in both boreholes and assets repositories. If this can be guaranteed we should implement **fast-langdetect** as our main method for language detection.

This issue is dependent on the completion of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/310.

### LINKED ISSUE #74: Use boreholes-dataextraction as a dependency in assets-dataextraction
Install the package that was created in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/310

Use the code provided by boreholes-dataextraction for language detection, instead of the current implementation.

When done remove fastext from Docker, then try with th new language.

christianabbet:
Closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/74 and closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/63

# Summary

Replaces the local language detection logic with the shared implementation from `swissgeol-boreholes-dataextraction`. Removes the FastText dependency and related Docker setup, resulting in a lighter and more maintainable project.

# Change
* Updated the project to use Python 3.11, aligning with swissgeol-boreholes-dataextraction.
* Synchronized dependency versions (e.g., Pydantic, boto3) to ensure compatibility.
* Removed all FastText dependencies and model files from the Docker image.
* Added new tests to validate model behavior for unsupported or unknown languages.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "swissgeol-assets-dataextraction"
 version = "0.1.0"
 description = "Python project for geological document analysis."
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11,<3.14"
 
 # Production-only dependencies for API runtime
 dependencies = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,11 +11,12 @@ requires-python = ">=3.10"
 
 # Production-only dependencies for API runtime
 dependencies = [
-    "boto3>=1.39.14,<2.0.0",
-    "boto3-stubs[s3]==1.40.53",
-    "boto3-stubs[essentials]==1.40.53",
-    "boto3-stubs[textract]==1.40.53",
-    "fasttext-predict>=0.9.2.4,<0.10.0.0",
+    "swissgeol-boreholes-dataextraction @ https://github.com/swisstopo/swissgeol-boreholes-dataextraction/releases/download/v1.0.124/swissgeol_boreholes_dataextraction-1.0.124-py3-none-any.whl",
+    "boto3==1.40.12",
+    "fastapi==0.116.1",
+    "boto3-stubs[s3]>=1.40.12,<2.0.0",
+    "boto3-stubs[essentials]>=1.40.12,<2.0.0",
+    "boto3-stubs[textract]>=1.40.12,<2.0.0",
     "fastapi[standard]==0.116.1",
     "joblib>=1.5.1,<2.0.0",
     "numpy>=2.0.0,<3.0.0",
@@ -23,8 +24,8 @@ dependencies = [
     "pandas>=2.3.1,<3.0.0",
     "pillow>=11.3.0,<12.0.0",
     "pymupdf==1.26.3",
-    "pydantic==2.12.2",
-    "pydantic-settings==2.11.0",
+    "pydantic==2.11.7",
+    "pydantic_settings==2.10.1",
     "pyyaml>=6.0.2,<7.0.0",
     "python-dotenv>=1.1.1,<2.0.0",
     "regex==2025.7.34",
--- a/src/language_detection/detect_language.py
+++ b/src/language_detection/detect_language.py
@@ -1,26 +1,19 @@
 import logging
 import math
-import os
 import re
 
-import fasttext
 import pymupdf
+import swissgeol_doc_processing as swiss
 from dotenv import load_dotenv
 
 load_dotenv()
-
-# Global model for detector empty if not used
-detector = None
-model_path = os.getenv("FASTTEXT_MODEL_PATH")
-if not model_path or not os.path.isfile(model_path):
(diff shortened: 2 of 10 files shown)


src/language_detection/detect_language.py
```diff
@@ -56,99 +49,90 @@ def extract_cleaned_text(page: pymupdf.Page) -> tuple[str, int]:
...
+def track_metadata_language(
+    lang: str,
     word_count: int,
     is_frontpage: bool,
     page_number: int,
     scores: dict[str, float],
     long_counts: dict[str, int],
 ) -> str | None:
-    """Selects metadata language and updates aggregated score trackers.
+    """Track metadata language and updates aggregated score trackers.
```
TicaGit, line 111: I think the return is always empty (None, wrong type hint)

TicaGit (commented):
LGTM

pyproject.toml
```diff
@@ -7,33 +7,34 @@ name = "swissgeol-assets-dataextraction"
...
     "joblib>=1.5.1,<2.0.0",
     "numpy>=2.0.0,<3.0.0",
     "opencv-python>=4.12.0.88,<5.0.0.0",
     "pandas>=2.3.1,<3.0.0",
     "pillow>=11.3.0,<12.0.0",
     "pymupdf==1.26.3",
-    "pydantic==2.12.2",
-    "pydantic-settings==2.11.0",
+    "pydantic==2.11.7",
+    "pydantic_settings==2.10.1",
```
stijnvermeeren-swisstopo, line 28: downgraded for compatibility with swissgeol-boreholes-dataextraction

src/language_detection/detect_language.py
```diff
@@ -56,99 +49,90 @@ def extract_cleaned_text(page: pymupdf.Page) -> tuple[str, int]:
...
+    if not text or not text.strip():
+        return None
 
     try:
-        labels, scores = detector.predict(text.lower(), k=5)
-        return [(label.replace("__label__", ""), score) for label, score in zip(labels, scores, strict=False)]
+        # If detected language is none of supported, return none
+        return swiss.language_detection.detect_language_of_text(
+            text=text.lower(), default_language=None, supported_languages=SUPPORTED_LANGUAGES
+        )
```
stijnvermeeren-swisstopo: The types of this method specify that `default_language` needs to be a string, but None seems to work as well.

src/language_detection/detect_language.py
```diff
@@ -1,26 +1,19 @@
 import logging
 import math
-import os
 import re
 
-import fasttext
 import pymupdf
+import swissgeol_doc_processing as swiss
```
stijnvermeeren-swisstopo: I'm not sure if I would abbreviate this to "swiss", can be a little confusing if you don't study the imports in detail...

stijnvermeeren-swisstopo (approved):
In theory, we could have exposed the language detection probabilities in the swissgeol-boreholes-dataextraction package, so that there would not have been a need to change the logic here. However, I don't think we ever had a very clear benefit from the thresholds anyway. so I'm ok with removing them for now.

I just added a few comments, but none of them are a blocker to merging this PR, so I'm already approving as well.

PR #77, opened 2025-11-18
feat/issue-68/pipeline-test-silent-fail

### LINKED ISSUE #68: Pipline tests silently fail and report success
## Issue 
While implementing https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/66, I noticed that tests were always successful, even though they should not be.

I realized this because `collecting tests/test_api_mapping.py` broke in https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/57 (because the `parse_predicted_class` is actually versionized with `parse_predicted_class_v1`), which is reflected in the run's results (https://github.com/swisstopo/swissgeol-assets-dataextraction/actions/runs/18467594659/job/52613250141). However, the pipeline still reported success.

Furthermore, since https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/61, the tests are no longer executed at all, and fail with (https://github.com/swisstopo/swissgeol-assets-dataextraction/actions/runs/18591674912/job/53007831889):
```
Run pytest --junitxml=pytest.xml --cov-report=term-missing:skip-covered --cov=src tests/ | tee pytest-coverage.txt
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov-report=term-missing:skip-covered --cov=src
  inifile: /home/runner/work/swissgeol-assets-dataextraction/swissgeol-assets-dataextraction/pyproject.toml
  rootdir: /home/runner/work/swissgeol-assets-dataextraction/swissgeol-assets-dataextraction
```
Again, they do so silently, so the test pipeline always reports success. 

## Likely cause

The new error message is likely due to https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/61/commits/dbe4d0605a2c8fb92b0d5318282052494c8c87d3, where `pytest-cov` was moved to `dev` dependencies, which are currently not installed (https://github.com/swisstopo/swissgeol-assets-dataextraction/blob/develop/.github/workflows/pytest.yml only installs `.test` dependencies).

## Proposed changes

* Test pipeline should report failure if the tests fail
* Test pipeline should execute properly by using `pytest-cov`

christianabbet:
Closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/68

# Overview

Change pycov to test package during CI install. Updating the test function in `tests/test_api_mapping.py` as it was pointing to a function that changed name making test fail.


pyproject.toml
```diff
@@ -42,13 +42,12 @@ dependencies = [
 dev = [
     "matplotlib>=3.10.3,<4.0.0",
     "mlflow>=3.3.1,<4.0.0",
-    "pytest-cov>=6.2.1,<7.0.0",
-    "shap==0.49.1",
```
stijnvermeeren-swisstopo, line 46: this is still needed by `src/models/treebased/model_explanation.py`, no?

stijnvermeeren-swisstopo (changes requested):
Unit test success/failure is now correctly reported again. However, the risk remains that if the dependency `pytest-cov` ever goes missing again, then the unit tests will once again fail silently. I would prefer to also modify `.github/workflows/pytest.yml` so that the status code of the `pytest` execution is reported correctly. I suspect that the reason why this is not yet the case, is related to the piped `tee` command (see e.g. https://unix.stackexchange.com/a/73180).

christianabbet:
The test now clearly fails. The test in tests/test_api_mapping.py was updated as it was silently failing after PR https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/78#.

stijnvermeeren-swisstopo (approved):

PR #78, opened 2025-11-19
type hint enum to see from the api

### LINKED ISSUE #72: Return possible classification values as Enum in OpenAPI specification
Feedback from https://github.com/swisstopo/swissgeol-assets-suite/issues/717#issuecomment-3484890504:
> Allenfalls könntet ihr bei euch aufnehmen, dass die möglichen Werte auch als Enum in der OpenAPI Specification ausgegeben werden. Dann können wir diese nämlich auch generieren und haben immer eine konsistente, von eurem Service vorgegebene Definition. (Das drängt aber nicht - allenfalls für 2026?).

TicaGit:
Resolves #72 

This PR updates the PageClasses enum to inherit from str and implements a dynamically created Enum with PascalCase values for the classes. This Enum is then exposed in the API making the enum values visible in the FastAPI documentation at /v1/docs.

Here is the nested structure showing the CollectResponse object.
image

### DIFF
--- a/src/page_classes.py
+++ b/src/page_classes.py
@@ -3,7 +3,7 @@
 from enum import Enum
 
 
-class PageClasses(Enum):
+class PageClasses(str, Enum):
     """Enum for classifying pages into page types."""
 
     TEXT = "text"
--- a/api/app/v1/schemas.py
+++ b/api/app/v1/schemas.py
@@ -1,6 +1,18 @@
+from enum import Enum
+from typing import TypeAlias
+
 from pydantic import BaseModel, ConfigDict, Field
+from pydantic.alias_generators import to_pascal
 
 from src.page_classes import PageClasses
+
+# dynamically created Enum to expose PascalCase class names to the API.
+PascalPageClasses = Enum(
+    "PascalPageClasses",
+    {name: to_pascal(value) for name, value in PageClasses.__members__.items()},
+    type=str,
+)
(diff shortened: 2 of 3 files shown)


src/page_classes.py
```diff
@@ -3,7 +3,7 @@
 from enum import Enum
 
 
-class PageClasses(Enum):
```
stijnvermeeren-swisstopo, line 6: The v1 API uses PascalCase labels, but since we now use these directly, the labels have changed to snake_case in the API response. We must preserve API compatibility.

stijnvermeeren-swisstopo (changes requested):

src/page_classes.py
TicaGit: Then I think we’re forced to create a new enum, one that exposes the PascalCase values for the API. Do you see any alternative approach?

TicaGit: Renaming all the classes from snake_case to PascalCase felt too large a change to handle here, as the names appear throughout the code, tests, README, and YAML configuration files. If this is considered necessary, it should be addressed in a dedicated issue. At that time, the code could also be cleaned of all the related mappings and aliases, mainly in mapping.py and page_classes.py.

For now, I implemented a dynamically created Enum to avoid hardcoding the class names twice. I also had to cast it as a type alias to use it in type hints. It’s a bit unconventional, but it works.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #84, opened 2026-01-19
Return document structure API

### LINKED ISSUE #81: Return document structure (esp. if consecutive pages are contiguous or independent)
Currently, the page classification pipeline classifies every page idependently. Hence, it makes no difference between "pages 5 and 6 are one borehole profile that continuous over two pages" and "pages 5 and 6 are two separate borehole profiles", it just returns the class "borehole profile" for both pages without saying anything else. The same applies to all other classes.

The goal is to change this, and to group together consecutive pages that "belong together" in the output of the data extraction pipeline. For the hypothetical examples above, the outputs could look like this:
```
[
  {
    "start_page": 5,
    "end_page": 6,
    "class": "Borehole"
  }
]
```
resp.
```
[
  {
    "start_page": 5,
    "end_page": 5,
    "class": "Borehole"
  },
  {
    "start_page": 6,
    "end_page": 6,
    "class": "Borehole"
  }
]
```

In a next step, we would also like to extract additional metadata for each of these elements, such a the name of a borehole, the scale of a map, or the title of a text chapter. However, first, we should change the API response format and figure out a way how we can accurately detect contiguous pages, and metadata extraction can be added in a later issue.

The ultimate goal is to enable this feature in the assets.swissgeol.ch web application: https://github.com/swisstopo/swissgeol-assets-suite/issues/794

For borehole profiles, the swissgeol-boreholes-dataextraction project already contains some logic to decide if borehole profiles on consecutive pages are parts of a single borehole, or separate boreholes. However, for other page classes, we might have to invent a completely new logic to decide the contiguity.

In a first version, we don't need a very accurate logic for all page classes yet. As a fallback, we could always group together consecutive pages with the same class and the same language, and work on splitting these more accurately later on. But we need at least some sort of a plan on how we want to deal with all the different page classes in this regard, to ensure that we don't implement something now that will turn out to be not flexible enough later on.

christianabbet:
Closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/81#

# Description

We introduce a new response schema for API v2. Results are returned per entity rather than per page. Each entity is defined by a start page, end page, page class, and language. The schema is designed to be extensible, for example to include page names for frontend rendering.

# Example

According to [Seitenklassifikation: Metadaten anzeigen #794](https://github.com/swisstopo/swissgeol-assets-suite/issues/794#issue-3681003564), pages are grouped by language and class. The only difference from the reference example is that Table page is currently classified as Borehole by the model for `35016.pdf`.

```json
{
  "has_finished": true,
  "data": {
    "filename": "35016.pdf",
    "page_count": 12,
    "languages": [
      "de"
    ],
    "entities": [
      {
        "classification": "text",
        "page_start": 1,
        "page_end": 1,
        "language": "de"
      },
      {
        "page_start": 2,
        "page_end": 3,
        "language": "de"
      },
      {
        "classification": "text",
        "page_start": 4,
        "page_end": 5,
        "language": "de"
      },
      {
        "classification": "map",
        "page_start": 6,
        "page_end": 6,
        "language": "de"
      },
      {
        "page_start": 7,
        "page_end": 9,
        "language": "de"
      },
      {
        "classification": "diagram",
        "page_start": 10,
        "page_end": 12,
        "language": "de"
      }
    ]
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,9 +147,6 @@ The V1 version containes extended classes from v0 and Each page is categorized i
 
 
 #### Example Output (v2)
-# TODO: Add test functions
-# TODO: Check that files are deleted after usage
-# TODO: Check aggregation with large document 35016
 
 ```json
 {
@@ -173,13 +170,6 @@ The V1 version containes extended classes from v0 and Each page is categorized i
 	],
 }
 ```
-**V1 Notes**:
-- `filename`: The name of the processed PDF file.
-- `metadata`: metadata about the file.
-- `pages`: list of dictionaries containing:
-  - `predicted_class`: The class name of the predicted class (e.g. "Boreprofile"). All possible classes are listed above in the section "Classes".
-  - `page_number`: The page number (1-indexed).
-  - `page_metadata`: metadata about the current page.
 
 
 #### General Notes
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ The V1 version containes extended classes from v0 and Each page is categorized i
 
 
 ## Output Format
-`data/prediction.json` (if `-w`/`--write_result`) or returned as a Python object.
+
 #### Example Output (v0)
 ```json
 {
@@ -146,7 +146,43 @@ The V1 version containes extended classes from v0 and Each page is categorized i
   - `page_metadata`: metadata about the current page.
 
 
-**General Notes:**
+#### Example Output (v2)
+# TODO: Add test functions
+# TODO: Check that files are deleted after usage
+# TODO: Check aggregation with large document 35016
+
+```json
+{
+	"has_finished": true,
+	"data": [
+		{
+			"filename": "input.pdf",					// Name of the file
+			"page_count": 3,							// Number of pages
+			"languages": [								// Detected languages
+				"de"
+			],
+			"entities": [								// List of elements present in file
+				{
+					"classification": "boreprofile",	// Type of element (PageClasses)
+					"page_start": 1,					// Starting page
+					"page_end": 3,						// Ending page
+					"language": "de",			    	// Detected language
+				},
+			],
+		}
+	],
+}
+```
+**V1 Notes**:
+- `filename`: The name of the processed PDF file.
+- `metadata`: metadata about the file.
+- `pages`: list of dictionaries containing:
+  - `predicted_class`: The class name of the predicted class (e.g. "Boreprofile"). All possible classes are listed above in the section "Classes".
+  - `page_number`: The page number (1-indexed).
+  - `page_metadata`: metadata about the current page.
+
+
+#### General Notes
 
 - The classifier supports batch input of multiple reports.
 - Input must be preprocessed: PDFs should already have OCR.
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ The V1 version containes extended classes from v0 and Each page is categorized i
 
 #### Example Output (v2)
 
-```json
+```jsonc
 {
 	"has_finished": true,
 	"data":{
--- a/README.md
+++ b/README.md
@@ -151,23 +151,21 @@ The V1 version containes extended classes from v0 and Each page is categorized i
 ```json
 {
 	"has_finished": true,
-	"data": [
-		{
-			"filename": "input.pdf",					// Name of the file
-			"page_count": 3,							// Number of pages
-			"languages": [								// Detected languages
-				"de"
-			],
-			"entities": [								// List of elements present in file
-				{
-					"classification": "boreprofile",	// Type of element (PageClasses)
-					"page_start": 1,					// Starting page
-					"page_end": 3,						// Ending page
-					"language": "de",			    	// Detected language
-				},
-			],
-		}
-	],
+	"data":{
+		"filename": "input.pdf",					// Name of the file
+		"page_count": 3,							// Number of pages
+		"languages": [								// Detected languages
+			"de"
+		],
+		"entities": [								// List of elements present in file
+			{
+				"classification": "boreprofile",	// Type of element (PageClasses)
+				"page_start": 1,					// Starting page
+				"page_end": 3,						// Ending page
+				"language": "de"			    	// Detected language
+			}
+		]
+	}
 }
 ```
 
--- a/api/app/v2/schemas.py
+++ b/api/app/v2/schemas.py
@@ -1,0 +1,38 @@
+from pydantic import BaseModel, ConfigDict
+
+from src.page_structure import ProcessorDocumentEntities
+
+
+class CollectResponse(BaseModel):
+    """Response model for the collect request endpoint containing processing status and results."""
+
+    has_finished: bool
+    data: list[ProcessorDocumentEntities] | None
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "example": {
+                "has_finished": True,
+                "data": [
+                    {
+                        "filename": "input.pdf",
+                        "metadata": {"page_count": 3, "languages": ["de"]},
+                        "entities": [
+                            {
+                                "start_page": 1,
+                                "end_page": 3,
+                                "lang": "de",
+                                "classification": "boreprofile",
+                                "data": None,
+                            },
+                        ],
+                    }
+                ],
+            }
+        }
+    )
+
+    @classmethod
+    def create_response(cls, predictions: list[dict]):
+        """Currently the api only handles requests with one file, `predictions` is a list of only one element."""
+        return cls(has_finished=True, data=[ProcessorDocumentEntities.model_validate(pred) for pred in predictions])
--- a/api/app/v2/schemas.py
+++ b/api/app/v2/schemas.py
@@ -19,10 +19,12 @@ class CollectResponse(BaseModel):
                         "metadata": {"page_count": 3, "languages": ["de"]},
                         "entities": [
                             {
-                                "start_page": 1,
-                                "end_page": 3,
-                                "lang": "de",
                                 "classification": "boreprofile",
+                                "metadata": {
+                                    "page_start": 1,
+                                    "page_end": 3,
+                                    "language": "de",
+                                },
                                 "data": None,
                             },
                         ],
--- a/api/app/v2/schemas.py
+++ b/api/app/v2/schemas.py
@@ -7,32 +7,30 @@ class CollectResponse(BaseModel):
     """Response model for the collect request endpoint containing processing status and results."""
 
     has_finished: bool
-    data: list[ProcessorDocumentEntities] | None
+    data: ProcessorDocumentEntities | None
 
     model_config = ConfigDict(
         json_schema_extra={
             "example": {
                 "has_finished": True,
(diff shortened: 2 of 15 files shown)


stijnvermeeren-swisstopo:
@christianabbet I don't really see why we would put the classification one level higher up in the JSON structure as `page_start` and `page_end` , because to me, the page range is even more fundamental than the classification. (For example, you might be more interested in the language than in the classification, but regardless, the page range will always be relevant).

At the moment, I don't see any real reason why we wouldn't put `classification`, `page_start`, `page_end` and `language` all at the same level.

Do we need to already add the `data` attribute, if it's always going to be empty at the moment? Do you think that we really need to create a new API version if we add an attribute to the JSON object later on, or would that essentially be a backward-compatible change for all practical purposes?

stijnvermeeren-swisstopo:
@christianabbet FYI I merged develop into this branch. Now `test_langs.py` is no longer showing up among the files that have changes.

christianabbet:
> @christianabbet I don't really see why we would put the classification one level higher up in the JSON structure as page_start and page_end , because to me, the page range is even more fundamental than the classification. (For example, you might be more interested in the language than in the classification, but regardless, the page range will always be relevant). At the moment, I don't see any real reason why we wouldn't put classification, page_start, page_end and language all at the same level.

It was mainly to align with the logic of API v1 where `filename` and `metadata` (containting `page_count` and `languages`) are at the top level. I would argue that if we put page_start, page_end and language one level higher then it would make sens to do the same with `page_count` and `languages`.

> Do we need to already add the data attribute, if it's always going to be empty at the moment? Do you think that we really need to create a new API version if we add an attribute to the JSON object later on, or would that essentially be a backward-compatible change for all practical purposes?

Indeed, we do not need to add the `data` attribute yet. I do not think we will need to create a new API version when adding data as it is a new field not the update of an existing one. So v2 will already work without data in it. It was just to "warn" user that something will come there in case they wanted to use it already.

As a result I would propose the following structure.  What do you think ?
```json
{
	"filename": "input.pdf",
	"page_count": 3,
	"languages": [
		"de"
	],
	"entities": [
		{
			"page_start": 1,
			"page_end": 3,
			"language": "de",
		},
	],
}
```

stijnvermeeren-swisstopo:
@christianabbet The simpler JSON structure proposed in your last comment seems preferable to me.

Looking at what is planned in the near future on the application side (https://github.com/swisstopo/swissgeol-assets-suite/issues/794), we are not really looking at extracting any rich data structures yet. The only thing that we will probably add in on the short term, would be an optional title/label for each section. Since this is probably an attribute that could be relevant regardless of the page classification, we might actually want to add it at the same level in the JSON schema, e.g.:
```
		{
			"page_start": 1,
			"page_end": 3,
			"language": "de",
			"title": "RB10",
		},
```

But as far as this ticket is concerned, would probably leave out the additional attribute for now. That would give us a little bit more time to reflect on the best naming (for example if someting like "label" might be more appropriate than "title").

README.md
```diff
@@ -146,7 +146,43 @@ The V1 version containes extended classes from v0 and Each page is categorized i
   - `page_metadata`: metadata about the current page.
 
 
-**General Notes:**
+#### Example Output (v2)
+# TODO: Add test functions
+# TODO: Check that files are deleted after usage
+# TODO: Check aggregation with large document 35016
```
stijnvermeeren-swisstopo: I guess these TODOs don't belong in the README?

README.md
```diff
@@ -146,7 +146,43 @@ The V1 version containes extended classes from v0 and Each page is categorized i
...
+					"page_start": 1,					// Starting page
+					"page_end": 3,						// Ending page
+					"language": "de",			    	// Detected language
+				},
+			],
+		}
+	],
+}
+```
+**V1 Notes**:
```
stijnvermeeren-swisstopo: Should this be "V2 notes" here?

README.md
letao: These can be removed I think

api/app/v2/schemas.py
```diff
@@ -0,0 +1,38 @@
...
+                    }
+                ],
+            }
+        }
+    )
+
+    @classmethod
+    def create_response(cls, predictions: list[dict]):
+        """Currently the api only handles requests with one file, `predictions` is a list of only one element."""
+        return cls(has_finished=True, data=[ProcessorDocumentEntities.model_validate(pred) for pred in predictions])
```
stijnvermeeren-swisstopo: Good point. Any reason not to just simplify the `data` attribute to be of type `ProcessorDocumentEntities | None`  instead of `list[ProcessorDocumentEntities] | None` ?

main.py
```diff
@@ -61,14 +67,139 @@ def flatten_dict(d, parent_key="", sep=".") -> dict:
     return dict(items)
 
 
+def group_consecutive(values: list[int]) -> list[list[int]]:
+    """Group sorted integers into consecutive sequences.
+
+    Args:
+        values: Sorted list of integers.
```
letao, line 74: We should check if we are still using a Sorted list of integers as input, if not we could update the Args

src/page_structure.py
```diff
@@ -25,16 +28,135 @@ class PageContext:
...
+
+    page: int
+    classification: PageClasses
+    metadata: ProcessorPageMetadata
+
+    @field_serializer("classification")
+    def classification_onehot(self, v: PageClasses, info: FieldSerializationInfo):
+        """Change type of classification representation based on context.
+
+        If legacy is provided throught context ({"legacy": True}), model returns onehoe encoding. This is
```
stijnvermeeren-swisstopo:
```suggestion
        If legacy is provided throught context ({"legacy": True}), model returns onehot encoding. This is
```

main.py
```diff
@@ -61,14 +67,139 @@ def flatten_dict(d, parent_key="", sep=".") -> dict:
...
+
+    # Processed PDFs
+    processor = PDFProcessor(classifier)
+    return processor.process_batch(pdf_files)
+
+
+def forward_document_entities(
+    documents: list[ProcessorDocument],
+) -> list[ProcessorDocumentEntities]:
+    """Convert classified docuemnts pages to entities.
```
letao: Typo in documents

src/page_structure.py
```diff
@@ -25,16 +28,135 @@ class PageContext:
     color_proportion: Counter | None = None
 
 
-class PageAnalysis:
-    """Stores the classification result for a single page."""
+class ProcessorPageMetadata(BaseModel):
+    """Processed pagee metadata."""
```
letao: Typo in page

src/pdf_processor.py
```diff
@@ -121,18 +119,22 @@ def process(self, file_path: Path) -> dict:
...
+                    )
                 )
 
         metadata = summarize_language_metadata(language_scores, long_page_counts, len(pages))
 
-        return {"filename": file_path.name, "metadata": metadata, "pages": pages}
+        return ProcessorDocument(
+            filename=file_path.name,
+            metadata=metadata,
+            pages=[p.model_dump() for p in pages],
```
letao: question about the line `pages=[p.model_dump() for p in pages]` why we couldn't just use `pages = pages`?

main.py
```diff
@@ -61,14 +67,139 @@ def flatten_dict(d, parent_key="", sep=".") -> dict:
...
+    documents_pages = forward_document(
+        pdf_files=pdf_files,
+        matching_params=matching_params,
+        model_path=model_path,
+        classifier_name=classifier_name,
+        explain_model=explain_model,
+    )
+
+    # Extract pages entities
+    documents_entities = forward_document_entities(documents=documents_pages)
```
stijnvermeeren-swisstopo: Maybe it keeps the code (method signatures) simpler if we only return `documents_pages` here, and then only compute the entities later on when they are actually needed?

src/page_structure.py
```diff
@@ -25,16 +28,135 @@ class PageContext:
...
+    ) -> Generator[tuple[tuple[PageClasses, str | None], list[ProcessorPage]], None, None]:
+        """Group pages by classes and languages.
+
+        Yields:
+            Generator[tuple[tuple[PageClasses, str | None], list[ProcessorPage]], None, None]: Returns
+                a generator with grouped pages par class and language along with corresponding tags.
+        """
+
+        # Get detected classes for each page
+        def key_fn(x: PageClasses) -> tuple[PageClasses, str | None]:
```
letao: Should this not be ` key_fn(x: ProcessorPage) -> tuple[PageClasses, str | None]:`

letao (changes requested):
Solid work, the new proposal for the JSON structure makes a lot of sense to me logic is also sound. 

The only comments I have are some small typos and overlooked classes.

src/page_structure.py
stijnvermeeren-swisstopo:
```suggestion
        def key_fn(x: ProcessorPage) -> tuple[PageClasses, str | None]:
```

main.py
```diff
@@ -79,52 +210,53 @@ def main(
...
+    documents_pages, documents_entities = forward(
+        input_path=input_path,
+        matching_params=matching_params,
+        model_path=model_path,
+        classifier_name=classifier_name,
+        explain_model=explain_model,
+    )
 
-    # Save to JSON
+    # Check if data need to be saves
```
stijnvermeeren-swisstopo:
```suggestion
    # Check if data need to be saved
```

stijnvermeeren-swisstopo (changes requested):
Good work, some minor comments, with some overlap with those that Davis just posted.

README.md
```diff
@@ -146,7 +146,31 @@ The V1 version containes extended classes from v0 and Each page is categorized i
   - `page_metadata`: metadata about the current page.
 
 
-**General Notes:**
+#### Example Output (v2)
+
+```json
```
stijnvermeeren-swisstopo:
```suggestion
```jsonc
```
Apparently, this will prevent the comments from being highlighted as syntax errors.
See e.g. https://github.com/stephenberry/JSONC

main.py
```diff
@@ -78,53 +170,65 @@ def main(
...
+    # End mlflow tracking
     if mlflow_tracking:
         mlflow.end_run()
 
-    if not write_result:
-        return results
+    if not return_entities:
+        return documents_pages
+    else:
+        return forward_document_entities(documents=documents_pages)
```
stijnvermeeren-swisstopo, line 231: Personally, at this point, I would probably have kept the previous tuple return type, for the benefit of type-safety. However, since this is something that can be simplified anyway as soon as we can remove the old API versions, I'm happy to merge it like this as well.

stijnvermeeren-swisstopo (approved):
Two non-essential small comments, up to you if you still want to make these changes, or leave the PR as-is. I'm already approving either way.

main.py
christianabbet: If it is fine for you I'll keep it like this to avoid computing the `forward_document_entities` and alter API v1 behaviour in case there is an issue with it.

letao (approved):

PR #93, opened 2026-02-02
Add document title extraction

### LINKED ISSUE #87: Return optional title for every document section (e.g. borehole name)
Based on:
- The borehole name extraction https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/300
- The code sharing with the swissgeol-boreholes-dataextraction repository https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/82
- The new response structure of the assets data extraction API https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/81

...we could now add an optional field `title` for every identified section of the document. For most classes, the title could be left empty for now, but it would be good if we can already start by returning the identified borehole name (if any) for the "borehole profile" class.

With the minimal pipeline for borehole detection, we probably don't have accurate information to distinguish between a borehole that continues across several pages, or separate boreholes on each page.

Reasonable assumption for now:
- start a new "borehole" document section if there is a different borehole name detected.
- continue with the existing "borehole" document section if no borehole name is detected, or if the detected name is equal to the name detected from the previous page.

christianabbet:
closes #87 

# Description
 
Adding a new field to the API response with the entity title name. It is left empty for all classes except Borehole profiles. To avoid conflicts with the extraction package, two config files are renamed for this project.

# Limitation

* **Run all pipeline**: The current borehole extraction package requires running the complete pipeline to detect titles. There is no option for shallow title extraction alone. Consequently, this implementation executes the full detection process. This PR extends beyond the original issue scope. If shallow title extraction is needed in the future, the extraction package will require updates to support this functionality.

* **Missing page**: When pages are initially classified as borehole-related but subsequently missed during actual borehole detection, these missed pages are assigned to the preceding borehole entry. Note that in the example below, page `1` remains isolated since it does not follow any existing borehole entry.
  * Initial borehole spans: `[[2], [4], [6, 7]]`
  * Missed pages: `[1, 3, 5, 8, 9]`
  * Resulting spans: `[[1], [2, 3], [4, 5], [6, 7, 8, 9]]`

# Example

Example of API v1 / v2 ouptut for `35016.pdf` classification.

```json
# V1
{
  "has_finished": true,
  "data": [
    {
      "filename": "35016.pdf",
      "metadata": {
        "page_count": 12,
        "languages": [
          "de"
        ]
      },
      "pages": [
        {
          "predicted_class": "Text",
          "page_number": 1,
          "page_metadata": {
            "language": "de",
            "is_frontpage": true
          }
        },
        {
          "page_number": 2,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 3,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 4,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 5,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Map",
          "page_number": 6,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 7,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 8,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 9,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 10,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 11,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 12,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        }
      ]
    }
  ]
}
```

```json
#V2
{
  "has_finished": true,
  "data": {
    "filename": "35016.pdf",
    "page_count": 12,
    "languages": [
      "de"
    ],
    "entities": [
      {
        "classification": "text",
        "language": "de",
        "page_start": 1,
        "page_end": 1,
        "title": null
      },
      {
        "language": "de",
        "page_start": 2,
        "page_end": 3,
        "title": null
      },
      {
        "classification": "text",
        "language": "de",
        "page_start": 4,
        "page_end": 5,
        "title": null
      },
      {
        "classification": "map",
        "language": "de",
        "page_start": 6,
        "page_end": 6,
        "title": null
      },
      {
        "language": "de",
        "page_start": 7,
        "page_end": 7,
        "title": "RB 1"
      },
      {
        "language": "de",
        "page_start": 8,
        "page_end": 8,
        "title": "RB 10"
      },
      {
        "classification": "diagram",
        "language": "de",
        "page_start": 9,
        "page_end": 12,
        "title": null
      }
    ]
  }
}
```


src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,139 @@
...
+    # Define page range
+    pages_id = list(range(page_start, page_end + 1))
+
+    # Write file to temp location for inference
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Create related files
+        out_directory = Path(tmpdir)
+        path_input = Path(out_directory) / pdf_file.name
+        path_prediction = Path(out_directory) / (pdf_file.name + ".pred.json")
+        path_metadata = Path(out_directory) / (pdf_file.name + ".meta.json")
```
AgathaSchmidt: Isn't output directory already a path as defined in the line above?

main.py
```diff
@@ -108,37 +111,74 @@ def forward_document(
...
+                page_start=page_start,
+                page_end=page_end,
+                language=language,
+            )
+        ]
+
+
 def forward_document_entities(
     documents: list[ProcessorDocument],
+    pdf_files: list[Path],
```
stijnvermeeren-swisstopo: Can we make `file_path` a property of the `ProcessorDocument` class? That would remove the need to pass these as separate list parameters and then zip them together again.

main.py
```diff
@@ -228,7 +268,7 @@ def main(
     if not return_entities:
         return documents_pages
     else:
-        return forward_document_entities(documents=documents_pages)
+        return forward_document_entities(documents=documents_pages, pdf_files=pdf_files)
```
stijnvermeeren-swisstopo: The following is not really a new issue, but it would still be good to fix it, because it would make it much easier to test the behaviour of the title extraction...

Currently, when running the script with `-w`, the results that are written to the `predictions.json` file are always the unstructured document page, before `forward_document_entities` is called. It would be better to write the fully structured output that includes the titles to `predictions.json` (as long as `return_entities == True`).

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+import json
+import logging
+import tempfile
+from itertools import chain
+from pathlib import Path
+
+import fitz
+from extraction.features.predictions.overall_file_predictions import OverallFilePredictions
+from extraction.main import start_pipeline
+from fitz import Document
```
stijnvermeeren-swisstopo: import from `pymupdf` instead of the older `fitz` import.

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Args:
+        pdf_document (Document): PDF to split.
+        pages_id (list[int]): List of pages to extract (1-based).
+
+    Returns:
+        Document: Selected subset.
+    """
+    # Create a new PDF for the selected pages
+    select_pdf = fitz.open()
```
stijnvermeeren-swisstopo: same here

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Args:
+        entities (list[ProcessedEntities]): List of detected entities.
+        pages_id (list[int]): Pages to match.
+
+    Returns:
+        list[int]: List of undetected pages.
+    """
+    pages_covered = [list(range(entity.page_start, entity.page_end + 1)) for entity in entities]
+    pages_covered_flattened = list(chain.from_iterable(pages_covered))
```
stijnvermeeren-swisstopo: I would find a set comprehension with a double for loop more readable, i.e. something like
```
pages_covered = {
    page_number
    for range in range(entity.page_start, entity.page_end + 1)}
    for page_number in range
}
```

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+
+        start_pipeline(
+            input_directory=path_input,
+            ground_truth_path=None,
+            out_directory=out_directory,
+            predictions_path=path_prediction,
+            metadata_path=path_metadata,
+            skip_draw_predictions=True,
+            part="all",
+        )
```
stijnvermeeren-swisstopo: It would be cleaner if we could pass an in-memory file as input (rather than a file path) and receive the results back as a return value, rather than having to read them back from a JSON file.

This might require some changes on the swissgeol-boreholes-dataextraction side though, or is it maybe already made possible by @letao 's changes in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/376 and https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/88 ?

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+            title=borehole.metadata.name.feature.name if borehole.metadata.name else None,
+        )
+        for borehole in prediction.file_predictions_list[0].borehole_predictions_list
+    ]
+
+    # Add dummy pages if any missed
+    pages_missed_id = _find_undetected_pages(entities, pages_id)
+    entities, pages_missed_id = _assign_trailing_pages(entities, pages_missed_id)
+
+    # Add empoty pages if still exists
```
stijnvermeeren-swisstopo:
```suggestion
    # Add an individual entity per page if there are still missing pages
```

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+            page_end=page_id,
+            language=lang,
+            title=None,
+        )
+        for page_id in pages_missed_id
+    ]
+
+    # Return page sorted entities
+    all_entities = entities + entities_missed
+    return sorted(all_entities, key=lambda x: x.page_start)
```
stijnvermeeren-swisstopo: Let's sort by `x.page_end` as well as a tie-breaker. I.e. a borehole that starts on page 5 and ends on page 6 should come after a borehole that starts AND ends on page 5.

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+            to any entity during detection.
+
+    Returns:
+        tuple[list[ProcessedEntities], list[int]]: A tuple containing:
+            - The updated list of entities with extended page ranges where applicable.
+            - The list of page IDs that could not be matched to any entity.
+    """
+    # Keep track of pages that were not matched
+    pages_not_matched_id: list[int] = []
+    for id_missed in pages_missed_id:
```
stijnvermeeren-swisstopo: Maybe we should ensure that `pages_missed_id` is a sorted list, to ensure more stable behaviour?

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,176 @@
...
+        pdf_file (Path): Path to pdf file.
+        page_start (int): Starting page (1-based).
+        page_end (int): Ending page (1-based).
+        lang (str): Detected language.
+
+    Returns:
+        list[ProcessedEntities]: List of boreprofile as entities.
+    """
+    # Define page range
+    pages_id = list(range(page_start, page_end + 1))
```
stijnvermeeren-swisstopo: I would use the name `page_numbers` instead of `pages_id`, seems clearer. Similar for `pages_missed_id` etc.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (changes requested):
We should also make sure to disable MLFlow tracking for Borehole data extraction when this is triggered from the assets pipeline. Otherwise, the MLFlow dashboard gets polluted, and even worse, I sometimes run into errors such as _"Exception: Run with UUID d4d69f9891644ebbbaa06a4a4458f4b2 is already active. To start a new run, first end the current run with mlflow.end_run(). To start a nested run, call start_run with nested=True"_

stijnvermeeren-swisstopo:
I did a number of tests on various files from the Zurich, Thurgau and Deepwells datasets. Overall, I'm very satisfied with the results. The behaviour for how boreholes are grouped/split and labelled seems to be quite robust already, and will definitely be useful to the end-users.

What worries me a little though, is that we currently don't have a good way to run such tests more frequently / in a more automated way. We have benchmarks for the classification of individual pages, but not for the overall document structure. Do you have any ideas on how we could add such tests, to ensure that we don't accidentally break the good behaviour in the future? Implementing that would be beyond the score of this issue though, more something for a follow-up issue.

src/boreprofile/entity_parser.py
christianabbet: It should be part of the new PR https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/377. The function used to takes as input a file and return the prediction. I edited the function to support streams as well.

If it makes sense, we can delay this PR and wait for the merge to implement the in-memory approach

christianabbet:
> We should also make sure to disable MLFlow tracking for Borehole data extraction when this is triggered from the assets pipeline. Otherwise, the MLFlow dashboard gets polluted, and even worse, I sometimes run into errors such as _"Exception: Run with UUID d4d69f9891644ebbbaa06a4a4458f4b2 is already active. To start a new run, first end the current run with mlflow.end_run(). To start a nested run, call start_run with nested=True"_

I'm not sure yet how to handle this properly. The main issue is that both repos use the same environment variable name for MLflow tracking. One solution would be to rename the variable in the Assets repo to avoid the collision. Another option would be to enforce the envirnoement variable to FALSE when running the function.  This has to be done before imports.

>  What worries me a little though, is that we currently don't have a good way to run such tests more frequently / in a more automated way. We have benchmarks for the classification of individual pages, but not for the overall document structure. Do you have any ideas on how we could add such tests, to ensure that we don't accidentally break the good behaviour in the future? Implementing that would be beyond the score of this issue though, more something for a follow-up issue.

I'm still exploring the best approach. One option is to include a reference document in the examples that serves as test input. Essentially a concatenation of different page types with predefined expected outputs. However, this approach has some drawbacks, if we modify the classification model later, we'd need to update the reference document.

stijnvermeeren-swisstopo:
@christianabbet

> I'm not sure yet how to handle this properly. The main issue is that both repos use the same environment variable name for MLflow tracking. One solution would be to rename the variable in the Assets repo to avoid the collision. Another option would be to enforce the envirnoement variable to FALSE when running the function. This has to be done before imports.

Another possible solution: can we split the `start_pipeline` method on the boreholes side into two separate methods, one that has all the "bells and whistles" related to MLFlow, evaluations, analytics, visualisations, etc. And then another one, that can be called here, that has the core code for obtaining predictions from a given file, without all of those bells and whistles, i.e. more or less this part of the code: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/79819b35c05a392ee8b2245c0b4ca87ae35ece69/src/extraction/runner.py#L252-L347

It feels like this would be a cleaner solution, and it would be good to split this very long method anyway, but it would require some changes on the boreholes repository first.

> I'm still exploring the best approach. One option is to include a reference document in the examples that serves as test input. Essentially a concatenation of different page types with predefined expected outputs. However, this approach has some drawbacks, if we modify the classification model later, we'd need to update the reference document.

Sounds like a reasonable approach. Maybe it's a little harsh to make the build fail as soon as something in the classification/data extraction for the example document changes. For example, it could happen that we've improved the results for many other documents, but the example document happens to be one of the edge cases where a new method does not give perfect accuracy. But hopefully, that will happen rarely enough that this will not be a real problem.

I'd be happy to give this suggestion a try. Could you create an issue for it?

stijnvermeeren-swisstopo:
@christianabbet

> I'm not sure yet how to handle this properly. The main issue is that both repos use the same environment variable name for MLflow tracking. One solution would be to rename the variable in the Assets repo to avoid the collision. Another option would be to enforce the envirnoement variable to FALSE when running the function. This has to be done before imports.

Another possible solution: can we split the `start_pipeline` method on the boreholes side into two separate methods, one that has all the "bells and whistles" related to MLFlow, evaluations, analytics, visualisations, etc. And then another one, that can be called here, that has the core code for obtaining predictions from a given file, without all of those bells and whistles, i.e. more or less this part of the code: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/79819b35c05a392ee8b2245c0b4ca87ae35ece69/src/extraction/runner.py#L252-L347

It feels like this would be a cleaner solution, and it would be good to split this very long method anyway, but it would require some changes on the boreholes repository first.

> I'm still exploring the best approach. One option is to include a reference document in the examples that serves as test input. Essentially a concatenation of different page types with predefined expected outputs. However, this approach has some drawbacks, if we modify the classification model later, we'd need to update the reference document.

Sounds like a reasonable approach. Maybe it's a little harsh to make the build fail as soon as something in the classification/data extraction for the example document changes. For example, it could happen that we've improved the results for many other documents, but the example document happens to be one of the edge cases where a new method does not give perfect accuracy. But hopefully, that will happen rarely enough that this will not be a real problem.

I'd be happy to give this suggestion a try. Could you create an issue for it?

christianabbet:
@stijnvermeeren-swisstopo Thanks for the answer. I'll create two issues. One focused on decoupling MLflow from our extraction logic, and another for the additional testing.

src/page_structure.py
```diff
@@ -95,13 +96,15 @@ class ProcessorDocument(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     filename: str
+    path: Path
```
stijnvermeeren-swisstopo: Unfortunately, this now conflicts with the `-w` (write results to `predictions.json` file) option. I get an error "TypeError: Object of type PosixPath is not JSON serializable" in this case.

The underlying reason is that this `ProcessorDocument` class is used for two purposes, on the one hand as input for creating the document entities (where we need the `path`), and on the other hand as data to be write to `predictions.json` file (where we actually don't care about the `path` anymore).

A quick fix (that also sort-of addresses the comment https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/93#discussion_r2763206956) would be to say that we actually don't care about `predictions.json` anymore when `return_entities == False`. Then we could move the writing of the `prediction.json`, and the error would disappear (since we would only need JSON serialization for the class `ProcessorDocumentEntities` and no longer for `ProcessorDocument`). What do you think?
```python
    if not return_entities:
        return documents_pages
    else:
        entities = forward_document_entities(documents=documents_pages)

        # Check if data need to be saved
        if write_result:
            output_file = Path("data") / "prediction.json"
            output_file.parent.mkdir(parents=True, exist_ok=True)
            output_file.write_text(
                json.dumps([r.model_dump() for r in entities], indent=4),
                encoding="utf-8",
            )

        return entities
```

src/boreprofile/entity_parser.py
```diff
@@ -0,0 +1,155 @@
...
+            page_end=page_number,
+            language=lang,
+            title=None,
+        )
+        for page_number in page_numbers_missed
+    ]
+
+    # Return page sorted entities
+    all_entities = entities + entities_missed
+    return sorted(all_entities, key=lambda x: x.page_end)
```
stijnvermeeren-swisstopo: > Let's sort by `x.page_end` as well as a tie-breaker. I.e. a borehole that starts on page 5 and ends on page 6 should come after a borehole that starts AND ends on page 5.

Sorry, maybe this was not very clearly phrased, but I meant to use `page_end` as a tie-breaker only, i.e. it should only be relevant for entities with an equal value for `page_start`, which should still be used as the primary sorting criterion.

stijnvermeeren-swisstopo (changes requested):

src/page_structure.py
christianabbet: Thank you for the catch! I missed that sorry. I found that you can hise fields in pydantic. So there are used within the class but excluded from dump. What do you think? If you don't like it I can go for the other option as well.

```python
    filename: str
    path: Path = Field(exclude=True)
    metadata: ProcessorDocumentMetadata
```

stijnvermeeren-swisstopo: I'd be happy with that solution as well.

stijnvermeeren-swisstopo: However, it would still be good to also address the comment linked above (write actual document structure to predictions.json when that option is active, instead of always just writing the plain page classification results).

main.py
```diff
@@ -274,6 +265,14 @@ def main(
     if not return_entities:
         return documents_pages
     else:
+        # Check if data need to be saved
+        if write_result:
+            output_file = Path("data") / "prediction.json"
+            output_file.parent.mkdir(parents=True, exist_ok=True)
+            output_file.write_text(
+                json.dumps([r.model_dump() for r in documents_pages], indent=4),
```
stijnvermeeren-swisstopo: We should write the results of `forward_document_entities(documents=documents_pages)`, not `document_pages`.

stijnvermeeren-swisstopo (changes requested):

main.py
```diff
@@ -236,7 +265,18 @@ def main(
     if not return_entities:
         return documents_pages
     else:
-        return forward_document_entities(documents=documents_pages)
+        # Check if data need to be saved
+        entities = forward_document_entities(documents=documents_pages)
+
+        # Check if data need to be saved
```
stijnvermeeren-swisstopo: comment duplicated

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):

PR #95, opened 2026-02-10
Change behaviour of Title page detection

### LINKED ISSUE #89: Change behaviour of "TitlePage" class
The current class TitlePage does not add much value. Often, it is used for the first page of an article, the cover page of a map, etc. However, from a "document structure" point of view, it would make more sense to consider this cover page to be part of the article, map, etc, instead of being a separate part of the document.

We also have some logic for identifying a "front page", which is however currently returned as metadata and not as a page class of its own. These are pages that were created by swisstopo (or some predecessor) and contain metadata for the document. As such they are more separate from the rest of the document and it would make more sense to have a dedicated page class for them. There is only a limited number of different layouts for these pages (two examples below) and they are often titles "Belegblatt", "Auszug aus dem Titelverzeichnis" or similar.

Consequently, I would propose the following:
- The pages that are currently classified as TitlePage should receive a different relevant class
- The pages that are currently identified as "FrontPage" should become the new training data for the TitlePage class
- Remove the separate logic for identifying the "FrontPage" metadata attribute, and instead integrate the detection into the regular xgBoost classification.

Image

Image

### LINKED ISSUE #94: Dataextraction Service liefert unterschiedliche Resultate
In `assets`, we use Cypress for e2e tests. During a rather large refactor, https://github.com/swisstopo/swissgeol-assets-suite/pull/679, significant progress has been made, and another rework in https://github.com/swisstopo/swissgeol-assets-suite/pull/711 sought to align the local and pipeline e2e environments. This rework introduced a certain flakiness to our tests, where the local environment works perfectly fine, yet the pipeline sometimes fails or succeeds.

After several hotfixes (https://github.com/swisstopo/swissgeol-assets-suite/pull/839, https://github.com/swisstopo/swissgeol-assets-suite/pull/805, https://github.com/swisstopo/swissgeol-assets-suite/pull/801), we still have this flakiness. Given that the tests run fine on local environments, and only tests that interface with the data extraction fail, our next guess is that the data extraction service returns different results for the same PDFs depending on where it is run.

# Basic Setup of e2e tests

* e2e tests trigger normal application flow, backend infrastructure is handled via `docker-compose` (https://github.com/swisstopo/swissgeol-assets-suite/blob/develop/development/docker-compose.yaml)
* Before tests, a `seed` is run (https://github.com/swisstopo/swissgeol-assets-suite/blob/develop/apps/server-asset-sg/src/commands/fixtures-create.command.ts) that declaratively triggers the extraction flow; basically calling the dataextraction API with the PDFs
* There are expectations (https://github.com/swisstopo/swissgeol-assets-suite/blob/develop/libs/shared/v2/src/lib/fixtures/files.fixtures.ts) that are compared to what the extraction service delivers; differences are logged to the console
  * NB: This flow is required by the current architecture because we need to properly register both files and extraction content - see https://github.com/swisstopo/swissgeol-assets-suite/issues/680 for a follow up ticket to properly refactor this
* There are two files which are processed (https://github.com/swisstopo/swissgeol-assets-suite/tree/develop/libs/shared/v2/src/lib/fixtures/files)
 
# The issue

* When run **locally**, expectations and actual results match; and hence, the tests work fine.
* When run **in the pipeline**, expectations do not match, but they are not always the same:
  * Example 1: https://github.com/swisstopo/swissgeol-assets-suite/actions/runs/21718698909/job/62944958202
    * Expectation mismatch reported:
    ```2026-02-09T08:51:43.849Z     WARN   FixturesCreateCommand  Stored page classifications do not match the result expected for this fixture.  [file: "a2_haute-sorne_2D_light_2023_swisstopo_technical_note.pdf", expected: [[to: 5, from: 1, languages: ["en"], categories: ["t"]], [to: 8, from: 6, languages: ["en"], categories: ["u"]], [to: 10, from: 9, languages: ["en"], categories: ["t"]], [to: 12, from: 11, languages: ["en"], categories: ["d"]], [to: 13, from: 13, languages: ["en"], categories: ["t"]]], actual: [[to: 1, from: 1, languages: ["de"], categories: ["t"]], [to: 5, from: 2, languages: ["en"], categories: ["t"]], [to: 8, from: 6, languages: ["en"], categories: ["u"]], [to: 10, from: 9, languages: ["en"], categories: ["t"]], [to: 12, from: 11, languages: ["en"], categories: ["d"]], [to: 13, from: 13, languages: ["en"], categories: ["t"]]]]```
    * In sum, the service identifies the first page as being `de`, which leads to a fully different classification as opposed to the expected `en` for page 1-5
    * The e2e tests fail, as a consequence, because we expect `EN` in the GUI
  * Example 2: https://github.com/swisstopo/swissgeol-assets-suite/actions/runs/21718623193
    * Same file, but a different run; Expectation mismatch reported:
    * Expectation mismatch reported:
    ```2026-02-05T16:05:02.053Z     WARN   FixturesCreateCommand  Stored page classifications do not match the result expected for this fixture.  [file: "a1_sample-report.pdf", expected: [[to: 2, from: 1, languages: ["en"], categories: ["u"]], [to: 4, from: 3, languages: ["en"], categories: ["t"]], [to: 8, from: 5, languages: ["en"], categories: ["u"]], [to: 9, from: 9, languages: ["en"], categories: ["t"]], [to: 10, from: 10, languages: ["fr"], categories: ["u"]]], actual: [[to: 1, from: 1, languages: ["de"], categories: ["u"]], [to: 2, from: 2, languages: ["en"], categories: ["u"]], [to: 4, from: 3, languages: ["en"], categories: ["t"]], [to: 8, from: 5, languages: ["en"], categories: ["u"]], [to: 9, from: 9, languages: ["en"], categories: ["t"]], [to: 10, from: 10, languages: ["fr"], categories: ["u"]]]]```
    * Here is another german title cover and the expectations change due to different page ranges - but because it's another file, our e2e tests work since for the other file, the expectations did not mismatch.

# Tested so far

* Rebuilt everything locally to make sure we're using the same image version both in pipeline and local env
* Cleared cache for Github actions
* Rerun tests multiple times with no deterministic threshold for success

# Possible issues
* Race conditions in test environment?
  * _Maybe_, expectations are mixed up, but we'd need to add more logging. However, it is hard to debug since there does not appear to be some kind of logic to it
* Caching issues
  * _Maybe_, the successful runs (from above) are happening by accident if caches from previous runs are reused - needs further testing
* Non deterministic results from the dataextraction service
  * If the same PDF returns different values (e.g. due to probabilistic matching), our  approach does not work anyways; but then it would be strange for it to be working locally _always_
* Different locales
  * _Maybe_ if the system locale of the Python process can influence the API service or dataextraction, this would explain the differences, but it would not explain why it fails inconsistently.

@stijnvermeeren-swisstopo I'd be happy if you could judge whether the issue _could_ be due to the extraction service. I will do some tests on our part as well; maybe we'll find the solution together.

christianabbet:
Closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/89# 
Closes https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/94# 

# Description

The model `models/stable/model.joblib` has been updated to include `section_header` as a new class. This class contains all examples from the previous dataset that were labeled as `title_page`. The new `title_page` class is now based on 100+ examples of actual title pages.

The dataset on S3 has been updated accordingly in `single_pages`, and the training/validation splits are available under `single_pages_splits`.

A VSCode launch template for training has been added. The document `CLASSIFICATION.md` provides details about the dataset, training, and classification of the model.

NB: After classification, `section_header` pages are assigned the category of the following page. If it is the last page in the document, it is mapped to `title_page`.

# Performances

| Class			  | Precision | Recall | F1-score |
|-----------------|----------:|-------:|---------:|
| boreprofile     |		 96.7 |   87.9 | 	 92.1 |
| diagram   	  |	     84.6 |   84.6 | 	 84.6 |
| geo_profile     |		 55.6 |   71.4 | 	 62.5 |
| map   		  |		 63.6 |   80.8 | 	 71.2 |
| section_header  |		 64.7 |   73.3 | 	 68.8 |
| table   		  |		 90.9 |   83.3 | 	 87.0 |
| text   		  |		 84.4 |   88.4 | 	 86.4 |
| title_page   	  |		 95.0 |   95.0 | 	 95.0 |
| unknown   	  |		 57.9 |   39.3 | 	 46.8 |
| Overall (macro) |		 77.0 |   78.2 |     77.1 |

# Example

Ouptut version for API v1 and v2

```json
#V1
{
  "has_finished": true,
  "data": [
    {
      "filename": "35016.pdf",
      "metadata": {
        "page_count": 12,
        "languages": [
          "de"
        ]
      },
      "pages": [
        {
          "page_number": 1,
          "page_metadata": {
            "language": "de",
            "is_frontpage": true
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 2,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 3,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 4,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 5,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Map",
          "page_number": 6,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 7,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 8,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 9,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 10,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 11,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 12,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        }
      ]
    }
  ]
}
```

```json
#V2
{
  "has_finished": true,
  "data": {
    "filename": "35016.pdf",
    "page_count": 12,
    "languages": [
      "de"
    ],
    "entities": [
      {
        "language": "de",
        "page_start": 1,
        "page_end": 1,
        "title": null
      },
      {
        "classification": "text",
        "language": "de",
        "page_start": 2,
        "page_end": 5,
        "title": null
      },
      {
        "classification": "map",
        "language": "de",
        "page_start": 6,
        "page_end": 6,
        "title": null
      },
      {
        "language": "de",
        "page_start": 7,
        "page_end": 7,
        "title": "RB 1"
      },
      {
        "language": "de",
        "page_start": 8,
        "page_end": 8,
        "title": "RB 10"
      },
      {
        "classification": "diagram",
        "language": "de",
        "page_start": 9,
        "page_end": 12,
        "title": null
      }
    ]
  }
}
```


src/page_classes.py
```diff
@@ -1,21 +1,23 @@
...
+    GEO_PROFILE = "geo_profile"
+    MAP = "map"
+    SECTION_HEADER = "section_header"
     TABLE = "table"
+    TEXT = "text"
+    TITLE_PAGE = "title_page"
     UNKNOWN = "unknown"
 
 
+# TODO: remove all the below after cleaning by DO
```
stijnvermeeren-swisstopo: Can this already be done now? I assume that "cleaning by DO" refers to https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/97, which has been merged now.

.gitignore
```diff
@@ -21,8 +21,8 @@ venv/
 minio
 
 # IDE config
-.idea/
```
stijnvermeeren-swisstopo, line 24: let's keep `.idea` in here, for people using PyCharm/IntelliJ

main.py
```diff
@@ -111,7 +112,52 @@ def forward_document(
...
+
+    # Reverse iteration to update classes
+    for i in range(len(classes))[::-1]:
+        # Check if current page is section header
+        if classes[i] != PageClasses.SECTION_HEADER:
+            continue
+
+        # If page is last, put to tile, otherwise set to class of next page
+        if i == len(classes) - 1:
+            document.pages[i].classification = PageClasses.TITLE_PAGE
```
stijnvermeeren-swisstopo: Some AI code review might help to spot typos such as "tile" <> "title". ;)

Also, I would prefer using the class "unknown" in this case, and to restrict "title_page" to those pages that are actually classified as such.

api/app/v1/schemas.py
```diff
@@ -110,6 +114,9 @@ def predicted_class(classification: PageClasses) -> PascalPageClasses:
     The values of the dict are the sting representation of each class in the PageClasses enum.
     """
     try:
+        # Merge section header and section headers
+        if classification == PageClasses.SECTION_HEADER:
+            classification = PageClasses.TITLE_PAGE
```
stijnvermeeren-swisstopo: I assume that this is to avoid the "section_header" class popping up in the V1 API endpoints? Or is there a different purpose for this? Maybe a comment to clarify would be good.

stijnvermeeren-swisstopo (changes requested):
A few initial comments. I haven't done any actual testing yet; I'll get to that tomorrow.

api/app/v1/schemas.py
christianabbet: It is indeed to avoid any leak. I just realized that it make more sense to put it to `unkown` as well to be consistent with your previous comment.

CLASSIFICATION.md
```diff
@@ -0,0 +1,97 @@
+# Classification
+
+To run classification, development package needs to be installed and MLFLow tracking activated.
```
stijnvermeeren-swisstopo: In what sense is MLFlow tracking necessary to run classification?

CLASSIFICATION.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+## Data
+
+The dataset is stored in the S3 bucket `stijnvermeeren-assets-data`, under the `single_pages/` folder. It contains categorized subfolders per class.
```
stijnvermeeren-swisstopo: While it's certainly not wrong to mention our swisstopo dataset and the accuracies on that data, I think that the primary audience for documentation on GitHub should be third-party developers. So we should be more clear about the fact that this data is not publicly available, and be clear about how people can construct their own datasets.

main.py
```diff
@@ -113,7 +113,52 @@ def forward_document(
 
     # Processed PDFs
     processor = PDFProcessor(classifier)
-    return processor.process_batch(pdf_files)
+    documents_pages = processor.process_batch(pdf_files)
+
+    # Affect section pages to page after
```
stijnvermeeren-swisstopo: "Affect" seems to be a strange choice of word here. What extractly do you want to describe here?

stijnvermeeren-swisstopo (changes requested):
The behaviour of the new "title page" classes and the _latent_ "section header" class is satisfactory. I'm pleasantly surprised by the accuracy of the title page classification, even without adding the older "frontpage"/"belegblatt" logic from `src/language_detection/pages_to_ignore.py` as a new feature for the classifier. I'm happy to merge it like this, though I would still create a follow-up ticket for deciding if we can be even more accurate in identifying title pages by somehow recycling this older code.

I also discovered that the ground truth data for the "unknown" class seems to contain a number of pages that should actually be classified as "title page". I would say that fixing though is out-of-scope for this issue, however. I think that the "unknown" class in general is a bit chaotic, and I will create a follow-up issue to re-think how we deal with this class.

My remaining comments on the PR are limited to documentation. Regarding the comments on `CLASSIFICATION.md`, please check with @letao , it might make more sense if he addresses those comments in #92 .

CLASSIFICATION.md
christianabbet: Not really, it's just the way the code is structured (`train.py`). Most of the logic is tightly coupled with MLflow, so I think actually made things easier when the code was created. That said, if you'd like to decouple the two, I'm happy to do that.

```
    if not mlflow_tracking:
        raise RuntimeError("MLflow tracking is disabled. Set MLFLOW_TRACKING=True in .env to enable it.")
```

stijnvermeeren-swisstopo: The point is that the documentation says "to **run** classification", which seems to imply not just training, but also inference. And for inference, MLFlow is optional, right? So I would clarify in the documentation that this sentence applies to training only, and then everything is fine.

stijnvermeeren-swisstopo (approved):
LGTM! :)

PR #102, opened 2026-02-24
Extraction of title from "section header" pages

### LINKED ISSUE #98: Extraction of title from "section header" pages
Extract the actual title from any page classified as "section header", and return it as the title for the corresponding document entity. 
For example, if the following pages are classified as text, then return the extracted title as the title for that text section.
If there are several consecutive pages that are classified as "section header", then use the title from the first page from which a title could be extracted.

For the title extraction itself, I would start with some relatively simple heuristic rules, e.g. looking at text size and positioning. Though if anyone has a good idea for another method that's not too complex and might be more robust, then I'm very open to try that out as well.

We don't have good ground truth for this yet, so I would propose the following:
- Create an initial baseline version
- Extract titles from all section header pages in the single pages dataset, and initially store the extracted title as the ground truth.
- Manually check/correct all of the extracted titles and add missing ones.
- Use that data to start benchmarking further improvements to the title extraction.

There's no code for benchmarking metadata such as titles yet, so that will have to be added as part of this issue as well.

christianabbet:
Close #98 

# Description

In this PR we include a title estimation for pages that are labeled as section headers. The ground truth in `data/gt_single_pages.json` is updated with title entries. To extract the title GT, we used the Pixtral model and a predefined prompt. The Pixtral interface was refactored so that it can be used in the future for other tasks. For title detection we send the image along with the extracted text. Pixtral seems to have poor OCR capability.

Moreover, we started some refactoring of the evaluation process. We opened a new issue for future cleanup (https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/103)

# Performance

The title is estimated based on multiple criteria and achieves **29.05%** F1-score. For each criterion, the cumulative score is indicated.

* Font size: Titles are more likely to have a larger font (18.99%)
* Horizontality: The title should begin in the left half of the page (20.11%)
* Verticality: The title should begin within the top 75% of the page vertically (23.46%)
* Length: The title should be at least 5 characters long (24.58%)
* Alpha: The title should consist mostly of alphabetic characters, with few numbers (25.70%)
* Highness: The title is more likely to appear near the top of the page (**29.05%**)

# API response

```jsonc
// API V2
{
  "has_finished": true,
  "data": {
    "filename": "35016.pdf",
    "page_count": 12,
    "languages": [
      "de"
    ],
    "entities": [
      {
        "language": "de",
        "page_start": 1,
        "page_end": 1,
        "title": null
      },
      {
        "classification": "text",
        "language": "de",
        "page_start": 2,
        "page_end": 5,
        "title": "Boden Wasser Luft"
      },
      {
        "classification": "map",
        "language": "de",
        "page_start": 6,
        "page_end": 6,
        "title": null
      },
      {
        "language": "de",
        "page_start": 7,
        "page_end": 7,
        "title": "RB 1"
      },
      {
        "language": "de",
        "page_start": 8,
        "page_end": 8,
        "title": "RB 10"
      },
      {
        "classification": "diagram",
        "language": "de",
        "page_start": 9,
        "page_end": 12,
        "title": null
      }
    ]
  }
}
```

```jsonc
// API V1
{
  "has_finished": true,
  "data": [
    {
      "filename": "35016.pdf",
      "metadata": {
        "page_count": 12,
        "languages": [
          "de"
        ]
      },
      "pages": [
        {
          "page_number": 1,
          "page_metadata": {
            "language": "de",
            "is_frontpage": true
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 2,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 3,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 4,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Text",
          "page_number": 5,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Map",
          "page_number": 6,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 7,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "page_number": 8,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 9,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 10,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 11,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        },
        {
          "predicted_class": "Diagram",
          "page_number": 12,
          "page_metadata": {
            "language": "de",
            "is_frontpage": false
          }
        }
      ]
    }
  ]
}
```


src/classifiers/pixtral_classifier.py
```diff
@@ -18,16 +19,95 @@
...
+    Attributes:
+        text (str | None): Text content, or None if image is provided.
+        image (PixtralImage | None): Image content, or None if text is provided.
+    """
+
+    text: str | None = None
+    image: PixtralImage | None = None
+
+    @model_validator(mode="after")
+    def at_least_one_field(self):
```
AgathaSchmidt, line 56: I can't find where this function is used

src/entity/utils.py
```diff
@@ -0,0 +1,25 @@
+"""Base utils for entity extraction."""
```
AgathaSchmidt: why do we need a special utils.py file? 
Currently the function is used in only one file (borehole_parser.py)

AgathaSchmidt (changes requested):

src/classifiers/pixtral_classifier.py
christianabbet: It's all good. It's called by the decorator `@model_validator` after object is created by pydantic

src/evaluation.py
```diff
@@ -21,66 +24,74 @@
...
+        text_gt (str): Ground truth text.
+        text_pred (str): Predicted text.
+        r_error (float, optional): Accepted relative error. Defaults to 0.25.
+
+    Returns:
+        bool: True if both texts are considered close to each other.
+    """
+    text_gt = standardize_text(text_gt)
+    text_pred = standardize_text(text_pred)
+    return distance(text_gt, text_pred) / max(1, len(text_gt)) < r_error
```
stijnvermeeren-swisstopo: might be simpler (and almost equivalent) to use the method https://rapidfuzz.github.io/Levenshtein/levenshtein.html#ratio here?

src/entity/titlepage_parser.py
```diff
@@ -0,0 +1,112 @@
...
+    the text of the highest-scoring candidate.
+
+    Args:
+        page (pymupdf.Page): The PDF page to analyse.
+
+    Returns:
+        str: Detected title for the page.
+    """
+    # Extract text block from page
+    extraction_context = extract_and_cache_page_data(page)
```
stijnvermeeren-swisstopo: How is this related to the `ExtractionContext` that is created as part of the `PageContext` in `pdf_process.py`. Can we somehow reduce it, to make the pipeline more computationally efficient?

Also, what kind of caching is happening in this method? The name says "cache", but I could not immediately see any caching happening when looking at the code.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
Code looks good, and the quality of the extracted titles is not perfect, but certainly good enough for a first version.

Some ideas for additional features that might help us identify titles more accurately:
- Is the text written in all-caps?
- Decrease score if the text includes words that are typical for authors and/or institutions (e.g. "école", "Bundesamt", etc.)
- Distance on the page to neighboring text (higher = better)
- Enforce consistent font size inside title text (kind-of the opposite of https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/108)

Additionally, it might be good to follow-up on the following technical debt:
- The borehole name is extracted for all files at the end of the pipeline, whereas section header title is extracted immediately during processing. This inconsistency is related to the ongoing support for the legacy V1 API, where the creating of document sections is not relevant, so we can leave it for now, but at the latest when we remove the V1 API, we should also work on making this more consistent.
- It would be useful to output document-level metrics and even page-level metrics/output to manually analyse and/or use in “diff tool”. -> https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/107

@christianabbet if you agree, then could you create follow-up tickets for these ideas (as long as they don't exist yet)?

christianabbet:
> @christianabbet if you agree, then could you create follow-up tickets for these ideas (as long as they don't exist yet)?

I created / updated the issue #107, #109, and #110. They should cover the additional features ideas, code refactoring for title/name detection, and the output of document/page-level metrics.

AgathaSchmidt (approved):

PR #106, opened 2026-03-09
Integration test: Document reference

### LINKED ISSUE #96: Add integration test using a reference document
Followup https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/93

# Description

Add a reference document to the examples that serves as test input. Essentially a concatenation of different page types with predefined expected outputs. This would allow us to verify that classification and extraction work end-to-end.

Keep in mind that the classification model may change over time. The test should be designed in a way that minimises the need for updates when that happens.

christianabbet:
Close https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/96#issue-3921968077

# Description

Add tests to verify the output coherence of the pipeline. This complements `.github/workflows/pipeline_run.yml`, which validates that the pipeline runs successfully but does not check the consistency of its output.

To perform these tests, a PDF is constructed from single pages covering all supported classes using the script `src/scripts/create_reference_document.py`. The resulting `reference_document.pdf` is fed through the pipeline. The tests then verify that:
* Page count matches the number and index of detected pages
* Entities fall within the document page range
* Document page classification is coherent with detected entities (classes match across input and output)


main.py
```diff
@@ -246,7 +246,7 @@ def main(
     write_result: bool = False,
     explain_model: bool = False,
     return_entities: bool = False,
-) -> list[ProcessorDocument] | list[ProcessorDocumentEntities]:
+) -> list[ProcessorDocument] | tuple[list[ProcessorDocument], list[ProcessorDocumentEntities]]:
```
stijnvermeeren-swisstopo: Wouldn't it be more robust to have a return type like `tuple[list[ProcessorDocument], list[ProcessorDocumentEntities] | None` ? Then you can avoid that the return type suddenly changes completely from a list to a tuple, just by switching a boolean parameter to the other value.

Not very critical though, since this will likely be simplified when we can deprecate the V1 API.

stijnvermeeren-swisstopo (approved):
One comment, but nothing critical, so I'm already approving as well.

main.py
christianabbet: I think it is a good point. It also makes the code more readable with a single return.

PR #111, opened 2026-03-19
Rethink Unknown class for classification

### LINKED ISSUE #101: Rethink handling of the "unknown" class in document classifier
# Description

During evaluation of the document classifier, it was discovered that the ground truth data for the "unknown" class contains a number of pages that should actually be classified as "title page". Examples of mislabelled pages include: `11613_1.pdf`, `18774_1.pdf`, `19115_1.pdf`, `19462_1.pdf`, `20993_1.pdf`, `28777_1.pdf`, and `35787_1.pdf`. There also appears to be some overlap / gray area between "unknown" and other classes more generally, suggesting the "unknown" class is somewhat chaotic and inconsistently defined.

# Task
Rather than treating "unknown" as a proper class with dedicated training data, we propose removing it as an explicit class and instead using a confidence threshold approach:
* Train the XGBoost classifier on the well-defined classes only (excluding "unknown").
* Extract a confidence score from the classifier for each prediction.
* If the confidence score falls below a defined threshold, return "unknown" instead of the top-scoring class.

christianabbet:
Close #101

# Description

We updated the page classification logic to treat the `unknown` class as an outlier class, meaning all samples labeled as `unknown` are considered out-of-distribution (OOD). The XGBoost model is trained on the remaining classes as usual. During evaluation, the confidence score of a new sample is compared against those from the training set. If it exceeds a defined threshold, the prediction is trusted, otherwise, it falls back to unknown. The thresholds are defined class based.

With this fix, we also updated the GT for few samples that were relabeled from unknown to title page. By doing so, the optimal hyper parameters changed.

The model submistted with the current PR is the `XGBoost`.

# Changes

Two approaches were explored for OOD threshold estimation:
* **Half-Normal**: Fits a half-normal distribution to model the variance in per-class confidence scores. The threshold is derived from the resulting confidence distribution. This approach does not use OOD samples for threshold estimation.
* **GMM (Gaussian Mixture Model)**: For each class, fits a two-component GMM. One Gaussian for the in-distribution confidence scores and one for the OOD samples of that class. The threshold is estimated as the probability of a point belonging to the in-distribution component.

Language prediction was non-deterministic in the previous swissgeol implementation when two languages were tied. Upgrading to the latest package resolves this for multiple files including: `20655_6`, `45004_28`, and `24652_1`.

# Results

The results below are obtained after hyperparameter fine-tuning. The current configuration file reflects the latest optimal values for each hyperparameter. 31 files were affected by class changes. Of these:
* 17 pages changed from a known class to `unknown`
* 8 pages changed from `unknown` to another class
* 6 pages changed from one known class to another

Out of the 31 affected files, 6 are now correctly classified: `35613_49`, `44179_151`, `1801_51`, `32987_9`, `32987_11`, and `17939_13`.

More than half of the errors occur when classifying the following classes:

* diagram: `8739_6`, `43283_25`, `32987_51`, `1677_51`
* map: `8739_3`, `4828_19`, `44179_95`, `10299_31`, `28777_11`
* section_header: `Berichte_NAB_08-025_EWS_3`, `39789_2`, `20170_214`, `1432_3`, `1860_3`
* text: `24279-24282_23`, `18605_5`, `44179_97`, `25076_160`

Upon visual inspection, no obvious reason could be found to explain why these files were classified as `unknown` rather than their expected classes.

| Classes        | XGBoost (old GT)   | XGBoost | XGBoostOOD |
|------------|---------|---------|------------|
| boreprofile    | 92.06  | 90.63  | 87.50   | 
| diagram        | 84.62  | 88.00  | 91.68   |
| geo_profile    | 62.50  | 62.50  | 50.00  | 
| map            | 71.19  | 70.18  | 68.85     |
| section_header | 68.75  | 83.87  | 75.00      |
| table          | 86.96  | 81.82 | 78.26  |
| text           | 86.36| 89.89   | 87.18     | 
| title_page     | 95.00  | 100.0  | 100.0      |   
| unknown        | 46.81 | 54.17  | 36.36     | 
| macro          | 77.14 | 80.12   | 75.25    | 

# Hyper parameters

```json
# XGBoost (old GT)  
{"max_depth": 6, "learning_rate": 0.1, "n_estimators": 600, "colsample_bytree": 1.0, "subsample":0.8}
# XGBoost
{"max_depth": 5, "learning_rate": 0.025, "n_estimators": 500, "colsample_bytree": 1.0, "subsample":0.6}
# XGBoostOOD
{"max_depth": 7, "learning_rate": 0.025, "n_estimators": 300, "colsample_bytree": 0.8, "subsample":0.6, "ood_mode": gmm, "ood_confidence": 0.05}
```

# What's next

As a follow-up, we recommend investigating additional features to improve OOD detection for `diagram`, `map`, `section_header`, and `text`. Addressing these could meaningfully improve overall classification performance. One promising direction would be to extract visual features automatically using a pretrained model (e.g. DINO - ViTs-16). Such a model can run on CPU and extract a rich set of features for classification, removing the need for handcrafted features and manual investigation. Furthermore, since some pages currently go through feature extraction steps whose runtime varies depending on page content, using a pretrained model would offer the additional benefit of a fixed inference time per page.


src/entity/borehole_parser.py
```diff
@@ -132,7 +132,7 @@ def document_to_boreprofiles(
             language=lang,
             title=borehole.metadata.name.feature.name if borehole.metadata.name else None,
         )
-        for borehole in prediction.borehole_predictions_list
+        for borehole in prediction.predictions.borehole_predictions_list
```
AgathaSchmidt, line 135: when I run python main.py -i data/single_pages_splits/validation -g data/gt_single_pages.json -w on the branch I get an error for this line. 
It works, with the old version of this code with 
`for borehole in prediction.borehole_predictions_list`
Do you have the same issue? 
Why do you call prediction.predictions?

AgathaSchmidt (commented):
The code looks very good. 
I like how you made the train_parallel.py file unneeded by implementing the parallel loader. 

I left one important comment regarding functionality. 

Further, I could not spot any flaws in the code so far. Maybe I will review again on Tuesday.

src/entity/borehole_parser.py
christianabbet: Have you tried to use `uv sync --extra ` to update the package ? It is because the stucture changed in the new package version.

AgathaSchmidt: Yes, thank you, you are right, now it is working well!

src/models/treebased/train.py
```diff
@@ -150,62 +205,86 @@ def main(config_path: str, out_directory: str, tuning: bool = False):
...
+        partial(load_data_and_labels_parallel, max_workers=max_workers)
+        if parallel
+        else load_data_and_labels_sequential
+    )
+
+    X_train, y_train, k_train = load_fn(train_folder, label_lookup)
+    X_val, y_val, k_val = load_fn(val_folder, label_lookup)
+
+    elapsed = time.time() - start_time
+    logger.info(f"\nFeature extraction completed in {elapsed:.1f}s ({len(X_train) + len(X_val)} pages)")
```
stijnvermeeren-swisstopo: why these newlines at the start of the log messages?

src/evaluation.py
```diff
@@ -178,26 +197,26 @@ def save_stats(stats_classification: dict, csv_path: Path) -> Path:
...
+        [
+            label,
+            s["true_positives"],
+            s["false_negatives"],
+            s["false_positives"],
+        ]
+        for label, s in stats_classification.items()
+    ]
+
+    return save_csv(content=header + content, csv_path=csv_path)
```
stijnvermeeren-swisstopo: This conflicts with @AgathaSchmidt's approach in https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/117/changes. I think I prefer her approach with the `csv.DictWriter` over this (overly?) generic `save_csv` method.

stijnvermeeren-swisstopo (changes requested):
It's not quite clear to me yet how the pages with class "unknown" in the ground truth are used during training (if they are make any difference at all?). Could you clarify?

I did not test the changes to the training and explanation code, but the better integration of parallel and sequential training certainly looks like an improvement.

src/evaluation.py
christianabbet: I agree! I will modify such that it align with her PR

christianabbet:
@stijnvermeeren-swisstopo I added the `DictWriter` from the other merged PR. It is still wrapped in a less generic function. I think it is still worth it as the code is used in 3 different functions. If you think we should remove it let me know.

In addition, you can now find an extensive description of the training process in `src/models/treebased/model.py` for the `XGBOODClassifier` class.

src/evaluation.py
```diff
@@ -225,13 +225,13 @@ def save_classification_stats(stats_classification: dict, csv_path: Path) -> Pat
...
-        header=["Label", "True_Positives", "False_Negatives", "False_Positives"],
+        header=["label", "true_Positives", "false_negatives", "false_positives"],
         contents=[
             {
-                "Label": label,
-                "True_Positives": s["true_positives"],
-                "False_Negatives": s["false_negatives"],
-                "False_Positives": s["false_positives"],
+                "label": label,
+                "true_positives": s["true_positives"],
```
stijnvermeeren-swisstopo, line 232:
```
ValueError: dict contains fields not in fieldnames: 'true_positives'
```

stijnvermeeren-swisstopo (changes requested):
As discussed on Wednesday, I'm now also tending towards keeping the model with `"unknown"` as a regular class during training. The `"unknown"` training samples, even though they are a rather heterogeneous bunch, seem to help to keep the other classes within reasonable bounds. For example, with the OOD approach, we notice that some pages with large figures or even with handwritten text are starting to be classified as `"map"`. This is most likely due to the fact that the curves lines on these pages look somewhat "map-like", and all the other possible classes are much further away, so there is nothing that prevents these inputs from being classified as `"map"` with a sufficiently high confidence.

I still believe that the OOD approach could potentially be valuable in the future, maybe after we have established a bit more consistency in how the possible classes are defined and how the training data is labelled. So we can merge the OOD code as an optional setting. But I agree that it seems best not to enable it by default for now.

christianabbet:
@stijnvermeeren-swisstopo Thank you for the feedback. I corrected the typo for the dict and set back the "XGBoost" as the default one (no OOD)

stijnvermeeren-swisstopo (approved):
LGTM!

PR #112, opened 2026-03-20
Adding keywords list to features

### LINKED ISSUE #99: Add predefined title page keywords as a feature for page classification
Followup of https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/95#pullrequestreview-3820107138

The current approach relies on multiple features for page classification. It would be interesting to add a new feature using the predefined list of keywords in `src/language_detection/pages_to_ignore.py.` These predefined keywords could help improve the classification performance of our page detection model expecially for title and section header.

```json
# Existing features
{
  "features": [
    "KeywordsList",  <- Exmaple of new feature
    "Words Per Line",
    "Word Density",
    "Mean Left",
    "Text Width",
    "Indent Std Dev",
    "Capitalization Ratio",
    "Num Map Keyword Lines",
    "Grid Line Length Sum",
    "Non Grid Line Length Sum",
    "Line Angle Entropy",
    "Line Score",
    "Num Geo Profile Keywords",
    "Num Unit Keyword",
    "Y Scale OK",
    "X Scale OK",
    "Num Valid Borehole Descriptions",
    "Num Strip Logs",
    "Num Tables",
    "Num Boreholes",
    "Num Good Sidebars",
    "Best Sidebar Score",
    "Num Long or Horizontal Lines",
    "Text Line Count"
  ]
}
```

AgathaSchmidt:
# Changes 

Adding Keywords List **Score** to feature list making it 24 features instead of 23. 
The keyword list score is computed as the highest fraction of matched keyword groups (each group requiring at least one substring match) across predefined title-page patterns (as defined in pages_to_ignore). This score is then added as an additional feature in the XGBoost training. 

Retrained an XGBoost model with the new 24 features. 

# Performance 

Data from classification_report_json 
<google-sheets-html-origin>
**Feature** | **main f1** | **24 feature f1** 
-- | -- | -- 
boreprofile |0.92  | 0.906
diagram | 0.846  |0.846
geo_profile | 0.625 | 0.667
map | 0.712 | 0.712
section_header | 0.688 | 0.727 
table | 0.8696  | 0.818
text | 0.864 | 0.874
title_page | 0.95  | 0.95
unknown | 0.468  | 0.458
accuracy | 0.771  | 0.773
weighted_avg | 0.783  | 0.783

## Side effects

<google-sheets-html-origin>
File | main | 24 feature model | gt
-- | -- | -- | -- 
24521_1 | de | fr | Detailbild der Doku with italian entries
28777_11 | geo_profile | unknown | map
440_02_1 | text | unknown | title_page
32987_51 | unknown  | text | diagram
32987_9 | table  | boreprofile| table


src/models/feature_engineering.py
```diff
@@ -125,10 +126,14 @@ def compute_text_features(
...
-        list: A list of 23 computed feature values for the page. If no text lines are found, returns a zero vector.
+        list: A list of 24 computed feature values for the page. If no text lines are found, returns a zero vector.
     """
+    """Computes 24 numerical features used for XGBoost page classification models."""
     if not lines:
-        return [0.0] * 23  # Handle empty pages (15 base + 8 borehole features)
+        return [0.0] * 24
+
+    page_text = page.get_text("text") or ""
+    keywords_list_score = max_title_page_keyword_score(page_text)
```
christianabbet: Can you use `lines` instead as in `get_map_features`? Just feed the list to the function and then you don't need this line `page.get_text("text") or ""` in the main function `compute_text_features`

models/stable/model_original.joblib
christianabbet, line 1: You can remove the original one. Just replace the `model.joblib` in the same folder

src/models/treebased/model.joblib
christianabbet, line 1: Do you know why we have the model in the src files ?

src/models/feature_engineering.py
```diff
@@ -125,10 +126,14 @@ def compute_text_features(
         extraction_context: Optional ExtractionContext for caching intermediate results.
 
     Returns:
-        list: A list of 23 computed feature values for the page. If no text lines are found, returns a zero vector.
+        list: A list of 24 computed feature values for the page. If no text lines are found, returns a zero vector.
     """
+    """Computes 24 numerical features used for XGBoost page classification models."""
```
christianabbet: Remove old function description

src/language_detection/pages_to_ignore.py
```diff
@@ -106,6 +106,36 @@
 }
 
 
+def normalize_text(text: str) -> str:
+    """Normalize page text before keyword matching."""
+    return " ".join(text.upper().split())
+
+
+def substring_match_ratio(text: str, substrings: list[list[str]]) -> float:
```
christianabbet, line 111: If we have time it would be nice to write a test function for this one. Just to make sure we keep it consitant. Also if you have time, can you add the full documentation ?

christianabbet (changes requested):

tests/test_substringmatching.py
```diff
@@ -0,0 +1,239 @@
...
+            ["SGM-DZ"],
+            ["BERICHTE"],
+            ["TABELLEN"],
+            ["NOTIZEN"],
+            ["DIAGRAMME"],
+            ["SEITEN"],
+            ["BEILAGEN"],
+        ],
+    }
+
```
christianabbet, line 33: you do not need to add the whole pattern, you can just test with a much ligther string I would say

tests/test_substringmatching.py
```diff
@@ -0,0 +1,239 @@
...
+    lines = _make_lines(text)
+
+    score = max_title_page_keyword_score(lines)
+
+    assert score == pytest.approx(8 / 9)
+
+
+def test_belegblatt_case(patched_title_patterns):
+    # 6491_1.pdf
+    text = """EB
```
christianabbet, line 64: Same here, maybe we don0t need the whole text to assess it ?

src/language_detection/pages_to_ignore.py
```diff
@@ -106,6 +108,63 @@
 }
 
 
+def normalize_text(text: str) -> str:
+    """Normalize page text before keyword matching."""
+    return " ".join(text.upper().split())
```
stijnvermeeren-swisstopo: Might the classifier not be even more accurate without normalizing the case? These kind of title pages are all based on standardised templates, so (apart from OCR errors) the template's words should always appear in the same way (uppercase/lowercase/...).

I also don't the point of doing `" ".join(text.upper().split())` and not just `text.upper()`.

stijnvermeeren-swisstopo (changes requested):

src/language_detection/pages_to_ignore.py
AgathaSchmidt: I have reported the influence of removing normalization in the PR. 
Most important differences are in title_page and section header in my opinion. 
What do you think? Which option is better?

stijnvermeeren-swisstopo: @AgathaSchmidt I think the changes in the numbers are too small to really say which one is better based on the evaluation. So I would go with the intuitive idea that the capitalization is indeed relevant in this case, and go ahead and remove the normalization.

AgathaSchmidt: Sorry, I made a mistake, I removed the normalization but still was upper casing all the candidates. 
I will re-run and report back. This will probably improve the results because by upper casing all the candidates but demanding exact match with title_page_substrings there were some keywords left out.

AgathaSchmidt: @stijnvermeeren-swisstopo Done

stijnvermeeren-swisstopo (approved):
LGTM!

christianabbet (approved):
LGTM

PR #115, opened 2026-03-30
Add uv dependency cooldown

### LINKED ISSUE #114: Add dependency cooldown for packages
Configure uv to only resolve package versions that are at least `N` days old, improving supply chain security by giving the community time to check new releases before they land in our environment.

christianabbet:
Closes #114 

Adds `exclude-newer = "7 days"` in `pyproject.toml`.

This prevents uv from resolving package versions published in the last 30 days, reducing supply chain risk by giving the community time to vet new releases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,9 @@ ignore = ["D102", "D105", "D107", "D100", "D103", "SIM108"]
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.uv]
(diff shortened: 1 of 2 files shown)


stijnvermeeren-swisstopo (approved):
LGMT!

One this that is not entirely clear is whether this would also block the newest releases of the swissgeol-boreholes-dataextraction package from being used in this repository for a whole week.
If that is the case, then we should add an exception to this rule.

However, we can happily merge this and test it in practice.

PR #116, opened 2026-04-07
Memory increase after pip to uv migration

### LINKED ISSUE #113: Memory increase after pip to uv migration
Following the migration from pip to uv, memory usage appears to have doubled from ~2GB to ~4GB. We want to investigate the root cause of this increase, as the uv migration itself should not have had a significant impact on memory consumption.

Areas to investigate:

* Package dependencies: Compare `requirements.txt` against `uv.lock` to identify any newly added or changed dependencies. Check whether uv resolved different package versions or pulled in additional transitive dependencies that were previously absent.
* Docker image size: Compare the GitHub-built images before and after the migration to detect any large difference in image size, which could indicate unexpected packages being bundled.
* Runtime memory profiling: Instrument the application to log memory usage as packages are loaded at startup. This should help pinpoint whether a specific package is responsible for the spike.

christianabbet:
Close #113 

# Description

We compare the memory usage from different build and settings

| Source | Version     | Image Size (GB) | Memory Idle (GB) |
|--------|-------------|-----------------|------------------|
| Local  | 1.1.1-dev14 | 2.29			 | 0.46				|
| GitHub | 1.1.1-dev14 | 3.37			 | 0.65				|
| Local  | 1.1.1-dev15 | 3.15			 | 1.67			 	|
| GitHub | 1.1.1-dev15 | 5.37 			 | 2.86				|
|  | | 	 | 	 	|
| Local  | 1.1.1-dev19 | 2.16		 | 0.46		 	|
| GitHub | 1.1.1-dev19 | 3.35 			 | 0.52				|

The version prior of UV merge (< 1.1.1-dev15) shows a lower memory usage.

# Investigation

We distinguish two different problems:
* Increase in RAM: The same issue as in Borehole, which leaked into assets. Fixing the problem in Borehole and applying the `--compile` flags resolves it.
* Increase in Size: The container was larger than expected due to a Docker issue involving a duplicated layer during the build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# --- Temporary stage used only during the build process
 FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS builder
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
@@ -10,13 +11,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 WORKDIR /app
 
-COPY pyproject.toml uv.lock ./
-
 # Install production dependencies only
-RUN uv sync --frozen --no-dev --no-install-project
+COPY pyproject.toml uv.lock ./
+RUN uv sync --frozen --no-dev --no-install-project --compile
 
-
-FROM python:3.13-slim-bookworm AS runtime
+# --- Final stage that becomes the actual shipped Docker image (starts fresh)
+FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS runtime
 
 ENV MLFLOW_TRACKING="False"
 ENV TMP_PATH=/tmp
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -10,13 +11,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-
 # Install production dependencies only
-RUN uv sync --frozen --no-dev --no-install-project
-
+COPY pyproject.toml uv.lock ./
+RUN uv sync --frozen --no-dev --no-install-project --compile
 
-FROM python:3.13-slim-bookworm AS runtime
+# --- Final stage that becomes the actual shipped Docker image (starts fresh)
+FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS runtime
```
letao: What would be the benefit of using this over the FROM python:3.13-slim-bookworm AS runtime that we had before?

pyproject.toml
```diff
@@ -63,7 +63,7 @@ all = [
 
 # ---- Configuration for UV
 [tool.uv.sources]
-swissgeol-boreholes-dataextraction = { url = "https://github.com/swisstopo/swissgeol-boreholes-dataextraction/releases/download/v1.0.144/swissgeol_boreholes_dataextraction-1.0.144-py3-none-any.whl" }
+swissgeol-boreholes-dataextraction = { url = "https://github.com/swisstopo/swissgeol-boreholes-dataextraction/releases/download/v1.0.0-dev4/swissgeol_boreholes_dataextraction-1.0.0.dev4-py3-none-any.whl" }
```
letao: I would suggest to wait until we can create a pre-release candidate here that we can reference so that we avoid the -dev4 issue. We could merge our develop branch into main on boreholes for that

letao (commented):
Looks good, only two minor comments from my side.

AgathaSchmidt (approved):
Looks good

letao (approved):

PR #117, opened 2026-04-07
add page level predictions

### LINKED ISSUE #107: Add Page-Level Predictions
# Description

Currently, the evaluation only reports false positives (FP), false negatives (FN), and true positives (TP) at a global level, aggregated across all documents. This makes it difficult to identify which specific documents are underperforming. It would be useful to output document-level metrics and even page-level metrics/output to manually analyse and/or use in a diff tool.

# Task

To improve observability, we'd like to add page-level prediction output, following the approach used in the Borehole repo (see `document_level_metrics.csv`). For example, the output should be a CSV file where each row represents a page, with columns for the page name, ground truth label, and predicted label. This will make it straightforward to identify and investigate failing files.

Here is an example of output (actual implementation might differ):

```csv
name, gt_title, pred_title,
1234.pdf, this is the GT title, this is the pred title
```

AgathaSchmidt:
Closes #107 

To address the additional page level output as proposed in #107 I created a new method save_page_level_predictions which 
writes one csv row per matched page (intersection of predictions and ground truth), with columns: 
- filename
- page
-  gt_title
-  pred_title
- gt_class
-  pred_class. 

Pages present in only one of the two dicts are skipped. 
The new file is saved  as `page_level_predictions`in the given output_dir which usually is evaluation, so that the csv is next ti the other evaluation metrics.

### DIFF
--- a/src/evaluation.py
+++ b/src/evaluation.py
@@ -136,24 +136,19 @@ def compute_title_stats(predictions: dict[str, DocumentPage], ground_truth: dict
     return {"title": stats}
 
 
-def compute_stats(
-    predictions: list[DocumentGroundTruth], ground_truths: list[DocumentGroundTruth]
-) -> tuple[dict, dict]:
+def compute_stats(predictions: dict[str, DocumentPage], ground_truths: dict[str, DocumentPage]) -> tuple[dict, dict]:
     """Compute classification and title extraction statistics against ground truth.
 
     Args:
-        predictions (list[DocumentGroundTruth]): Predicted document annotations.
-        ground_truths (list[DocumentGroundTruth]): Ground truth document annotations.
+        predictions (dict[str, DocumentPage]): Keyed predictions ('filename-page').
+        ground_truths (dict[str, DocumentPage]): Keyed ground truth ('filename-page').
 
     Returns:
         tuple[dict, dict]: A tuple of (classification_stats, title_stats), each as per-label
             confusion dictionaries.
     """
-    pred_keyed = groundtruth_doc_to_pages(predictions)
-    gt_keyed = groundtruth_doc_to_pages(ground_truths)
-
     # Evaluate on the intersection so we don't crash when pages are missing
-    pred_keys, gt_keys = set(pred_keyed.keys()), set(gt_keyed.keys())
+    pred_keys, gt_keys = set(predictions.keys()), set(ground_truths.keys())
 
     missing_in_pred = gt_keys - pred_keys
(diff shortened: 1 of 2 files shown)


src/evaluation.py
```diff
@@ -162,12 +157,44 @@ def compute_stats(
...
+                    "gt_title": gt_page.title or "",
+                    "pred_title": pred_page.title or "",
+                    "gt_class": next((k for k, v in gt_page.classification.items() if v), ""),
+                    "pred_class": next((k for k, v in pred_page.classification.items() if v), ""),
+                }
+            )
+    return csv_path
+
+
 def save_stats(stats_classification: dict, csv_path: Path) -> Path:
```
christianabbet: Maybe we can rename it `save_classification_stats`

src/evaluation.py
christianabbet: Do you mind refactoring this piece as well using the same logic as in `save_page_level_predictions` with `csv.DictWriter` I feel like it's more elegant how you wrote it. I'll then integrate it in my PR for Unkown class

christianabbet (commented):
LGTM, two minor comments, you can already ask Stijn for review

src/evaluation.py
```diff
@@ -162,13 +157,45 @@ def compute_stats(
...
+    Args:
+        predictions (dict[str, DocumentPage]): Keyed predictions ('filename-page').
+        ground_truth (dict[str, DocumentPage]): Keyed ground truth ('filename-page').
+        csv_path (Path): Destination path for the output CSV file.
+
+    Returns:
+        Path: The path to the written CSV file.
+    """
+    with open(csv_path, "w", newline="") as f:
+        writer = csv.DictWriter(f, fieldnames=["filename", "page", "gt_title", "pred_title", "gt_class", "pred_class"])
```
stijnvermeeren-swisstopo: I would prefer a different order, with the class columns coming before the title columns and the prediction coming before the ground truth. I.e.
```["filename", "page", "pred_class", "gt_class", "pred_title", "gt_title"]```

Like this, the information that is more complete, more concise and more likely to changes comes first.

stijnvermeeren-swisstopo (changes requested):
Works well! Just a small request regarding the column order in the CSV.

Also, the practice of creating a string key of the form `"{filename}-{page}"` and then splitting the key again into its individual components feels a little hacky, but since it was already there before, it's not really in the scope of this issue to improve it.

stijnvermeeren-swisstopo (approved):
LGTM!

Issue #108, opened 2026-03-11
Improve Text Block Aggregation


christianabbet:
Currently, the asset repo's aggregation method tends to over-segment text that should be grouped into a single block. Font size, in particular, could be given more weight during the grouping step to better consolidate related text fragments.

The following pages show concrete examples in section headers where text that should form a single block is incorrectly split into multiple ones:

* `1062_4`: "LOSINGER + CO AG BERN LABORATORIUM"
* `1630_80`: "Untern Gasterntale & ihre Bedeutung für die Weiterfuhrung"
* `2001_7`: "Rapport géologique sur le Glissement dans les Gorges de Court"
* `17415_62`: "Geologia e geotecnica del Sondaggio"
* `31042_2`: "GEOLOGISCHE GRUNDLAGEN DER HYDROMECHANISCHEN MODELLIERUNG"

Improving the aggregation logic for these cases would reduce fragmentation and lead to more accurate block detection across multilingual documents.

PR #121, opened 2026-04-15
Improve title detection section header

### LINKED ISSUE #109: Improve Title detection (Section Header)
# Description

Improve title identification for the `Section Header` class, as a follow-up to #102.

# Tasks

Several enhancements to increase title detection accuracy:

* **All-caps detection**: boost score if the text is written entirely in uppercase
* **Penalize author/institution keywords**: decrease score when the text contains words typical of author names or institutions (e.g. "école", "Bundesamt", "AG", "GmbH")
* **Distance to neighboring text**: take into account vertical/horizontal spacing to adjacent blocks. Greater isolation should increase the score.
* **Enforce consistent font size**: verify that font size is uniform across all spans within a candidate title block (check #108, as new block boundaries may affect this)

AgathaSchmidt:
Closes #109 
------------
Add properties: 
- all_caps 
- no_institution

Add method 
- _assign_isolation_scores --> better prediction for `34239_1.pdf`  (still wrong though)

Remove 
- horizontality --> this has the most influence on the changes --> detection of `28366_2.pdf`

Those are few of the proposed improvements from #109 , but actually they do not contribute much to the improvement. 
In the single pages validation dataset we have 14 section headers, out of which 12 are correctly classified as such (`1860_3.pdf` and `NTB_07-01_1.pdf` are classified as `unknown`). 
From those 12 section headers, 6 are exactly extracted in develop and on this branch one more is correctly detected (`28366_2.pdf`). 
Further, the prediction in `34239_1.pdf` improves, even though the correct header remains undetected, because the prediction now is more "header-like" : `Mont Terri Project` vs `SNHGS.SHGN ANDRA BGR ENRESA`
Screenshot 2026-04-15 at 16 43 58

Moreover, the main problem of the remaining section headers are multi-line headers which are not detected so that only one part/line is detected. I tried to solve this issue by merging those lines together. This improved the situation for `2001_2.pdf` and  `2001_7.pdf` from 

```
pred: über die Schuttrutschung in der
gt:   Geologisches Gutachten über die Schuttrutschung in der Klus von Court
pred: Rapport géologique
gt:   Rapport géologique sur le Glissement dans les Gorges de Court
```
to the correct GT headers. But unfortunately the GT is inconsistent there, because this multi line title is actually only “ Geologische Baugrunduntersuchungen” in the GT

`11380_2.pdf`
pred: Geologische Baugrunduntersuchungen für den Neubau des SBB-Aufnahmegebäudes Bahnhof Dietikon
gt:   Geologische Baugrunduntersuchungen
Screenshot 2026-04-15 at 16 48 21

**Therefore, I reversed the merging of the lines and this behaviour remains a problem, which is not possible to solve due to the inconsistent ground truth definitions.** 

All predictions 
<google-sheets-html-origin>
filename | gt | main | 109branch
-- | -- | -- | --
11380_2.pdf | section_header \| Geologische Baugrunduntersuchungen | section_header \| Geologische Baugrunduntersuchungen | section_header \| Geologische Baugrunduntersuchungen
1630_393.pdf | section_header \| Quartalbericht Nr. 10 | section_header \| Quartalbericht Nr. 10 | section_header \| Quartalbericht Nr. 10
16770_10.pdf | section_header \| Seismische Bestimmung der Felsoberfläche | section_header \| Seismische Bestimmung der Felsoberfläche | section_header \| Seismische Bestimmung der Felsoberfläche
1860_3.pdf | section_header \| ETUDE GEOELECTRIQUE | unknown | unknown
18774_2.pdf | section_header \| PTT - NEUBAU SANDACKER | section_header \| 2601 Archiv | section_header \| 2601 Archiv
19462_2.pdf | section_header \| NUTZBARMACHUNG DER GEOTHERMIE IM RAUME BADEN | section_header \| NUTZBARMACHUNG DER GEOTHERMIE IM RAUME BADEN | section_header \| NUTZBARMACHUNG DER GEOTHERMIE IM RAUME BADEN
2001_2.pdf | section_header \| Geologisches Gutachten über die Schuttrutschung in der Klus von Court | section_header \| Geologisches Gutachten | section_header \| über die Schuttrutschung in der
2001_7.pdf | section_header \| Rapport géologique sur le Glissement dans les Gorges de Court | section_header \| sur le Glissement dans les Gorges de Court | section_header \| Rapport géologique
28366_2.pdf | section_header \| INTERNER BERICHT 88-05 | section_header \| BELEnicht | section_header \| INTERNER BERICHT 88-05
34239_1.pdf | section_header \| GP Experiment: TDR Measurements in Borehole BGP-7 | section_header \| SNHGS.SHGN ANDRA BGR ENRESA | section_header \| Mont Terri Project
39789_2.pdf | section_header \| GISIHÜBELI - ZUCHWIL TUNNEL BIRCHI TUNNELPORTAL WEST | section_header \| GISIHUBELI - ZUCHWIL | section_header \| GISIHUBELI - ZUCHWIL
8583_2.pdf | section_header \| Geologische Baugrunduntersuchungen | section_header \| Geologische Baugrunduntersuchungen | section_header \| Geologische Baugrunduntersuchungen
Berichte_NAB_08-025_EWS_3.pdf | section_header \| Arbeitsbericht NAB 08-25 | section_header \| Arbeitsbericht NAB 08-25 | section_header \| Arbeitsbericht NAB 08-25
NTB_07-01_1.pdf | section_header \| Technical Report 07-01 | unknown | unknown


src/entity/titlepage_parser.py
```diff
@@ -1,55 +1,84 @@
...
+    """A normalized text block candidate for title scoring.
+
+    All positional attributes are expressed in page-relative coordinates
+    so that scores are comparable across pages of different sizes.
+    """
 
     text: str
     line_count: int
     rect: pymupdf.Rect
+    font_size: float
```
christianabbet: Remove variable not used

christianabbet (changes requested):
Overall the metric is not changing: Title: F1=43.75%, Precision=38.89%, Recall=50.00%

Still, I would argue to add tge changes even though they do not bring much in term of scoring. Maybe it would be worth discussing this in the meeting.

AgathaSchmidt:
@christianabbet Yes, the metrics are not changing and in the past it was discussed that this is expected. But we can discuss in the next Stand Up.

src/entity/titlepage_parser.py
```diff
@@ -60,52 +81,106 @@ def non_numericality(self) -> float:
...
+    def no_institution(self) -> float:
+        """Return a penalty factor when institution or author keywords are detected.
+
+        Checks for words typical of company names, government bodies, or research
+        institutions (e.g. 'AG', 'GmbH', 'Bundesamt', 'école').
+
+        Returns:
+            float: 0.2 if an institution keyword is found as a whole word, 1.0 otherwise.
+        """
+        words = set(re.split(r"[\s,;.:()\[\]/]+", self.text.lower()))
```
stijnvermeeren-swisstopo: Why not simply `words = set(re.findall(r"\w+", self.text.lower()))` `?

src/entity/titlepage_parser.py
```diff
@@ -60,52 +81,106 @@ def non_numericality(self) -> float:
...
+            * self.length
+            * self.non_numericality
+            * self.highness
+            * self.all_caps
+            * self.no_institution
+            * self.isolation
+        )
+
+
+def _assign_isolation_scores(candidates: list[TitleCandidateTextBlock]) -> None:
```
stijnvermeeren-swisstopo: It would seem more consistent with the other factors to implement this as a method on the `TitleCandidateTextBlock` class, that takes the list of all candidates as a parameter.

src/entity/titlepage_parser.py
```diff
@@ -60,52 +81,106 @@ def non_numericality(self) -> float:
...
+    """Set isolation on each candidate based on vertical gap to its nearest neighbour.
+
+    Scores range from 0.5 (block immediately adjacent to another) to 1.0
+    (block separated by ≥ 10 % of the normalized page height).
```
stijnvermeeren-swisstopo: Computing the gap relative to the font size of the candidate title might be a bit more robust than computing it relative to the page height. But I'd be happy to leave the initial implementation like this, and potentially improve on it later on.

stijnvermeeren-swisstopo (changes requested):
You write in the PR description that you managed to merge lines together to e.g. "Geologische Baugrunduntersuchungen für den Neubau des SBB-Aufnahmegebäudes Bahnhof Dietikon" for 11380_2.pdf. However, in the table of results, this is not the case. Could you please clarify?

Also, this PR should close #109, not #102 (start of PR description).

stijnvermeeren-swisstopo (changes requested):
To get some additional data to assess the accuracy, I also tested the behaviour on the section headers from the train subset.

A few observations from the extracted titles from those pages:
- Squaring the font size has a negative impact in some cases, e.g. `20170_2.pdf` where the stamp "VERTRAULICH" is now incorrectly extracted. To avoid extracting very short, very large text, maybe scoring the total area (width * height) of the text lines might be more accurate? (As long as we can then avoid extracting entire paragraphs of text as title.)
- The isolation score only considers vertical gaps. It would probably be more robust to include the actual 2D distance to the other blocks. For example, the two-column layout in `28134_3.pdf` leads to the actual title having a minimal rather than a maximal isolation score.

AgathaSchmidt:
> You write in the PR description that you managed to merge lines together to e.g. "Geologische Baugrunduntersuchungen für den Neubau des SBB-Aufnahmegebäudes Bahnhof Dietikon" for 11380_2.pdf. However, in the table of results, this is not the case. Could you please clarify?
Yes, this is not clear in the PR. 


I tried the merging of multiple title lines in commit [49c2af0](https://github.com/swisstopo/swissgeol-assets-`dataextraction/commit/49c2af052b7a3e0c70df96e46ef63e168b19538c) with `_merge_title_continuation` but then I reversed the changes since the ground truth is not consistent anyways and sometimes the full title is correct e.g. "Rapport géologique sur le Glissement dans les Gorges de Court" and sometimes only one line, e.g. "Geologische Baugrunduntersuchungen" . 
Due to the inconsistency of the gt I decided to stick to the current state as it is. 
Still, I wanted to share my insights in the PR. I will rephrase it so it is clear that it is not implemented in this version.

AgathaSchmidt:
> * Squaring the font size has a negative impact in some cases, e.g. `20170_2.pdf` where the stamp "VERTRAULICH" is now incorrectly extracted. To avoid extracting very short, very large text, maybe scoring the total area (width * height) of the text lines might be more accurate? (As long as we can then avoid extracting entire paragraphs of text as title.)

I could try something like this for sure! 
I expect that for pages like `34239_1`like  showed in the PR we would chose the random words at the top again. But lets see how the advantages and disadvantages balance each other out, I will give it a try  

Here are the results so far, for the super simple approach of just calculating: `self.rect.width * self.rect.height`

For the validation dataset we have only disadvantages. 
We are loosing the correct predictions from `11380_2`, `1630_393` and `2001_7` for the sake of a longer line or a combination of multiples lines. 
In `18774_2` and `34239_1` the new predictions are wrong, just as the old ones where, but from my perspective they make less sense, e.g. that `34239_1` behaved  as expected. 

On the other hand, the effect on the training dataset is great overall. 
I selected all documents of the class section header from training. 

before: 

=== Title extraction (gt_class == section_header, n=77) ===
Exact match:   14/77  (18.2%)
Partial match: 26/77  (33.8%)

now: 

=== Title extraction (gt_class == section_header, n=77) ===
Exact match:   17/77  (22.1%)
Partial match: 35/77  (45.5%) 

Exact matches gained: `1062_4.pdf`, `17415_59.pdf`, `20170_2.pdf`, `20655_3.pdf`, `27898_2.pdf`, `44044_5.pdf`

Exact matches lost:  `1630_231.pdf`, `24361_1.pdf`, `8583_3.pdf`

Partial matches gained (they usually extract too much): `16770_2.pdf` , `1677_3.pdf`, `18605_2.pdf`, `20907_1.pdf`,  `24279-24282_160.pdf`, `28777_2.pdf`, `31800_2.pdf`, `35613_4.pdf`, `4455_2.pdf`, `8583_3.pdf`


Do you think this is worth committing? 
Should I spend more time trying for a better solution which is more complex than simply multyplying height with width? 
Should I open a new ticket for that?

stijnvermeeren-swisstopo:
@AgathaSchmidt 
- How about average area per line (`self.rect.width * self.rect.height / max(self.line_count, 1)`)? I.e. a bit of a combination of the old and new formula. This seems to work really well on my tests. We should probably also rename the `font()` method. 
- The results for files like 11380_2 and 2001_7 will probably be improved when we deal with #108, so the accuracy on those pages is not critical to me yet.
- I don't see any implementation of the suggested 2D distance calculation for isolation score yet?

AgathaSchmidt:
The isolation score changes are helping for the case you mentioned `28134_3.pdf` but generally a drop of matches is noticeable. 

changing only the area calculation to the with average (`self.rect.width * self.rect.height / max(self.line_count, 1))`)
training dataset

=== Title extraction (gt_class == section_header, n=77) ===
  Exact match:   23/77  (29.9%)
  Partial match: 37/77  (48.1%)
  
  Adding isolation changes: 
  === Title extraction (gt_class == section_header, n=77) ===
  Exact match:   21/77  (27.3%)
  Partial match: 33/77  (42.9%)

stijnvermeeren-swisstopo (changes requested):
@AgathaSchmidt Thanks for the analysis. Considering those results, I agree with keeping the isolation based on vertical distance only for the time being.

I still think that a 2D-distance based isolation could be a good idea in the future, but as far as this issue is concerned, I think we already have a sufficiently significant improvement.

Could you maybe create a follow-up issue for the 2D distance, with the following two preconditions:
- Address #108 first
- Compute the distanced based on the original PDF coordinates, rather than on the normalized coordinates. Because with the normalized coordinates on a narrow page, the horizontal distances get inflated relative to the vertical distances, which is probably not beneficial.

For this PR:
- Revert the implementation of the isolation score to the one from before the last commit.
- I'm also not a big fan of having to call the `assign_isolation` method on every candidate before the scores are computed correctly. Could we modify that so that the value of `title_candidates` is either passed to the `TitleCandidateTextBlock` constructor, or to the `score` method?

src/entity/titlepage_parser.py
```diff
@@ -60,52 +79,101 @@ def non_numericality(self) -> float:
...
     """
-    # Extract text segments from page and convert to text blocks
     lines = extract_text_lines(page)
     text_blocks = create_text_blocks(lines)
 
-    # Create list of text candidates and return best
     title_candidates = [TitleCandidateTextBlock(text_block=text_block, rect=page.rect) for text_block in text_blocks]
-    title_candidates = sorted(title_candidates, key=lambda x: x.score, reverse=True)
-
+    title_candidates.sort(key=lambda x: x.score(title_candidates), reverse=True)
```
stijnvermeeren-swisstopo: I don't fully understand why, but it looks like it's a bad idea to use the list that is being sorted as a parameter in the `key` function. For some inputs (e.g. 20170_2.pdf) this seems to result in a canddiates list that is not actually correctly sorted (even though the `score` function actually looks to be independent of the ordering of this list...).

@AgathaSchmidt can you reproduce this, and if so, find a suitable fix?

stijnvermeeren-swisstopo (dismissed):

src/entity/titlepage_parser.py
stijnvermeeren-swisstopo: By the way, using `sorted` instead of `sort` seems to work better for me...

```
    title_candidates = sorted(title_candidates, key=lambda x: x.score(title_candidates), reverse=True)
```

christianabbet (approved):
LGTM

